PR #2, opened 2017-09-21
 Correctly handle different numbers of private subnets and NAT states 

eherot:
I'm not sure if this is exactly the right thing to be checking for but I know that without this or `enable_nat_gateway` set there are not likely to be any private subnets to check for. Perhaps we should use `var.enable_nat_gateway` instead?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -45,12 +45,12 @@ resource "aws_route" "public_internet_gateway" {
 # Private routes
 #################
 resource "aws_route_table" "private" {
-  count = "${length(var.azs)}"
+  count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.private_subnets)) : 0}"
 
   vpc_id           = "${aws_vpc.this.id}"
   propagating_vgws = ["${var.private_propagating_vgws}"]
 
-  tags = "${merge(var.tags, var.private_route_table_tags, map("Name", format("%s-private-%s", var.name, element(var.azs, count.index))))}"
+  tags = "${merge(var.tags, var.private_route_table_tags, map("Name", (var.single_nat_gateway ? "${var.name}-private" : format("%s-private-%s", var.name, element(var.azs, count.index)))))}"
 }
 
 ################
@@ -128,13 +128,13 @@ resource "aws_elasticache_subnet_group" "elasticache" {
 # NAT Gateway
 ##############
 resource "aws_eip" "nat" {
-  count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.azs)) : 0}"
+  count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.private_subnets)) : 0}"
 
   vpc = true
 }
 
 resource "aws_nat_gateway" "this" {
-  count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.azs)) : 0}"
+  count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.private_subnets)) : 0}"
 
   allocation_id = "${element(aws_eip.nat.*.id, (var.single_nat_gateway ? 0 : count.index))}"
   subnet_id     = "${element(aws_subnet.public.*.id, (var.single_nat_gateway ? 0 : count.index))}"
@@ -143,7 +143,7 @@ resource "aws_nat_gateway" "this" {
 }
 
 resource "aws_route" "private_nat_gateway" {
-  count = "${var.enable_nat_gateway ? length(var.azs) : 0}"
+  count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.private_subnets)) : 0}"
 
   route_table_id         = "${element(aws_route_table.private.*.id, count.index)}"
   destination_cidr_block = "0.0.0.0/0"
@@ -212,7 +212,7 @@ resource "aws_vpc_endpoint_route_table_association" "public_dynamodb" {
 # Route table association
 ##########################
 resource "aws_route_table_association" "private" {
-  count = "${length(var.private_subnets)}"
+  count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.private_subnets)) : 0}"
 
   subnet_id      = "${element(aws_subnet.private.*.id, count.index)}"
   route_table_id = "${element(aws_route_table.private.*.id, count.index)}"


antonbabenko:
Can you try this with example code which is inside the repository? Make sane, but failing values for example and share it, please.

eherot:
So here's a modification of the `simple-vpc` example with the NAT turned off and no private subnets defined:

```
module "vpc" {
  source = "../../"

  name = "no-nat-example"

  cidr = "10.0.0.0/16"

  azs            = ["eu-west-1a", "eu-west-1b", "eu-west-1c"]
  public_subnets = ["10.0.101.0/24", "10.0.102.0/24", "10.0.103.0/24"]
}
```
And here's an excerpt of what `terraform plan` outputs:
```
...

  + module.vpc.aws_route_table.private[0]
      id:                               <computed>
      route.#:                          <computed>
      tags.%:                           "1"
      tags.Name:                        "no-nat-example-private-eu-west-1a"
      vpc_id:                           "${aws_vpc.this.id}"

  + module.vpc.aws_route_table.private[1]
      id:                               <computed>
      route.#:                          <computed>
      tags.%:                           "1"
      tags.Name:                        "no-nat-example-private-eu-west-1b"
      vpc_id:                           "${aws_vpc.this.id}"

  + module.vpc.aws_route_table.private[2]
      id:                               <computed>
      route.#:                          <computed>
      tags.%:                           "1"
      tags.Name:                        "no-nat-example-private-eu-west-1c"
      vpc_id:                           "${aws_vpc.this.id}"
...
```
It also does something similarly irrational if you set `enable_nat_gateway = true` and `single_nat_gateway = true` and only define one private network. This is admittedly a bit of an edge case and probably a stupid way to set up a VPC but I tried to address it anyway because that is the situation I found at the company where I'm implementing this. Here's what it does:
```
...

  + module.vpc.aws_nat_gateway.this
      id:                               <computed>
      allocation_id:                    "${element(aws_eip.nat.*.id, (var.single_nat_gateway ? 0 : count.index))}"
      network_interface_id:             <computed>
      private_ip:                       <computed>
      public_ip:                        <computed>
      subnet_id:                        "${element(aws_subnet.public.*.id, (var.single_nat_gateway ? 0 : count.index))}"

  + module.vpc.aws_route.private_nat_gateway[0]
      id:                               <computed>
      destination_cidr_block:           "0.0.0.0/0"
      destination_prefix_list_id:       <computed>
      egress_only_gateway_id:           <computed>
      gateway_id:                       <computed>
      instance_id:                      <computed>
      instance_owner_id:                <computed>
      nat_gateway_id:                   "${element(aws_nat_gateway.this.*.id, count.index)}"
      network_interface_id:             <computed>
      origin:                           <computed>
      route_table_id:                   "${element(aws_route_table.private.*.id, count.index)}"
      state:                            <computed>

  + module.vpc.aws_route.private_nat_gateway[1]
      id:                               <computed>
      destination_cidr_block:           "0.0.0.0/0"
      destination_prefix_list_id:       <computed>
      egress_only_gateway_id:           <computed>
      gateway_id:                       <computed>
      instance_id:                      <computed>
      instance_owner_id:                <computed>
      nat_gateway_id:                   "${element(aws_nat_gateway.this.*.id, count.index)}"
      network_interface_id:             <computed>
      origin:                           <computed>
      route_table_id:                   "${element(aws_route_table.private.*.id, count.index)}"
      state:                            <computed>

  + module.vpc.aws_route.private_nat_gateway[2]
      id:                               <computed>
      destination_cidr_block:           "0.0.0.0/0"
      destination_prefix_list_id:       <computed>
      egress_only_gateway_id:           <computed>
      gateway_id:                       <computed>
      instance_id:                      <computed>
      instance_owner_id:                <computed>
      nat_gateway_id:                   "${element(aws_nat_gateway.this.*.id, count.index)}"
      network_interface_id:             <computed>
      origin:                           <computed>
      route_table_id:                   "${element(aws_route_table.private.*.id, count.index)}"
      state:                            <computed>

...

  + module.vpc.aws_route_table.private[0]
      id:                               <computed>
      route.#:                          <computed>
      tags.%:                           "1"
      tags.Name:                        "no-nat-example-private-eu-west-1a"
      vpc_id:                           "${aws_vpc.this.id}"

  + module.vpc.aws_route_table.private[1]
      id:                               <computed>
      route.#:                          <computed>
      tags.%:                           "1"
      tags.Name:                        "no-nat-example-private-eu-west-1b"
      vpc_id:                           "${aws_vpc.this.id}"

  + module.vpc.aws_route_table.private[2]
      id:                               <computed>
      route.#:                          <computed>
      tags.%:                           "1"
      tags.Name:                        "no-nat-example-private-eu-west-1c"
      vpc_id:                           "${aws_vpc.this.id}"
...
```

eherot:
Reworked (and squashed) this change a bit now that I understand the problem a little better. Also opted to add a new change: Use subnets instead of AZs to determine the count in a few places, since that is really the ultimate determiner of how many of these things we can make use of.

main.tf
```diff
@@ -45,12 +45,12 @@ resource "aws_route" "public_internet_gateway" {
...
 #################
 resource "aws_route_table" "private" {
-  count = "${length(var.azs)}"
+  count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.private_subnets)) : 0}"
 
   vpc_id           = "${aws_vpc.this.id}"
   propagating_vgws = ["${var.private_propagating_vgws}"]
 
-  tags = "${merge(var.tags, var.private_route_table_tags, map("Name", format("%s-private-%s", var.name, element(var.azs, count.index))))}"
+  tags = "${merge(var.tags, var.private_route_table_tags, map("Name", (var.single_nat_gateway ? "${var.name}-private" : format("%s-private-%s", var.name, element(var.azs, count.index)))))}"
```
antonbabenko: I think `tags` can be left as it was since there will be at least one route table element. The only change which is required is to use `var.private_subnets` instead of `var.azs`

main.tf
```diff
@@ -128,13 +128,13 @@ resource "aws_elasticache_subnet_group" "elasticache" {
 # NAT Gateway
 ##############
 resource "aws_eip" "nat" {
-  count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.azs)) : 0}"
+  count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.private_subnets)) : 0}"
```
antonbabenko: Such repetitive expressions you can simplify using locals:
```hcl
locals {
  number_of_nats =  "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.private_subnets)) : 0}"
}

resource "aws_eip" "nat" {
  count = "${local.number_of_nats}"
  ...
}
```

I am just not sure that all users of this module are using the version which supports `locals`.

Let's keep it for later.

eherot: Interesting! Hadn't seen those before...

main.tf
eherot: I was on the fence about this so I wouldn't mind going back if you prefer it the old way.

main.tf
antonbabenko: Actually, we should not use `locals`, because it is not possible to specify the version of module downloaded from the registry and it will break code for everyone who is using an older version of Terraform and using github instead of registry to download modules.

main.tf
lorengordon: Testing out this patch. I like this code as is. When creating a single nat gateway (and so a single private route table), the route table name tag doesn't need the az.

main.tf
```diff
@@ -212,7 +212,7 @@ resource "aws_vpc_endpoint_route_table_association" "public_dynamodb" {
 # Route table association
 ##########################
 resource "aws_route_table_association" "private" {
-  count = "${length(var.private_subnets)}"
+  count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.private_subnets)) : 0}"
```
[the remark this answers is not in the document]
lorengordon: Just tested this code, and it behaves the way I think was intended here:

```hcl
resource "aws_route_table_association" "private" {
  count = "${length(var.public_subnets)}"

  subnet_id      = "${element(aws_subnet.private.*.id, count.index)}"
  route_table_id = "${var.single_nat_gateway ? aws_route_table.private.id : element(aws_route_table.private.*.id, count.index)}"
}
```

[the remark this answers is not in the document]
lorengordon: Here's a diff applying this approach everywhere we were using `aws_route_table.private.*.id`:

```diff
diff --git a/main.tf b/main.tf
index 377d936..385dbf5 100644
--- a/main.tf
+++ b/main.tf
@@ -147,7 +147,7 @@ resource "aws_nat_gateway" "this" {
 resource "aws_route" "private_nat_gateway" {
   count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.private_subnets)) : 0}"

-  route_table_id         = "${element(aws_route_table.private.*.id, count.index)}"
+  route_table_id         = "${var.single_nat_gateway ? aws_route_table.private.id : element(aws_route_table.private.*.id, count.index)}"
   destination_cidr_block = "0.0.0.0/0"
   nat_gateway_id         = "${element(aws_nat_gateway.this.*.id, count.index)}"
 }
@@ -172,7 +172,7 @@ resource "aws_vpc_endpoint_route_table_association" "private_s3" {
   count = "${var.enable_s3_endpoint ? length(var.private_subnets) : 0}"

   vpc_endpoint_id = "${aws_vpc_endpoint.s3.id}"
-  route_table_id  = "${element(aws_route_table.private.*.id, count.index)}"
+  route_table_id  = "${var.single_nat_gateway ? aws_route_table.private.id : element(aws_route_table.private.*.id, count.index)}"
 }

 resource "aws_vpc_endpoint_route_table_association" "public_s3" {
@@ -202,7 +202,7 @@ resource "aws_vpc_endpoint_route_table_association" "private_dynamodb" {
   count = "${var.enable_dynamodb_endpoint ? length(var.private_subnets) : 0}"

   vpc_endpoint_id = "${aws_vpc_endpoint.dynamodb.id}"
-  route_table_id  = "${element(aws_route_table.private.*.id, count.index)}"
+  route_table_id  = "${var.single_nat_gateway ? aws_route_table.private.id : element(aws_route_table.private.*.id, count.index)}"
 }

 resource "aws_vpc_endpoint_route_table_association" "public_dynamodb" {
@@ -216,24 +216,24 @@ resource "aws_vpc_endpoint_route_table_association" "public_dynamodb" {
 # Route table association
 ##########################
 resource "aws_route_table_association" "private" {
-  count = "${var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.private_subnets)) : 0}"
+  count = "${length(var.public_subnets)}"

   subnet_id      = "${element(aws_subnet.private.*.id, count.index)}"
-  route_table_id = "${element(aws_route_table.private.*.id, count.index)}"
+  route_table_id = "${var.single_nat_gateway ? aws_route_table.private.id : element(aws_route_table.private.*.id, count.index)}"
 }

 resource "aws_route_table_association" "database" {
   count = "${length(var.database_subnets)}"

   subnet_id      = "${element(aws_subnet.database.*.id, count.index)}"
-  route_table_id = "${element(aws_route_table.private.*.id, count.index)}"
+  route_table_id = "${var.single_nat_gateway ? aws_route_table.private.id : element(aws_route_table.private.*.id, count.index)}"
 }

 resource "aws_route_table_association" "elasticache" {
   count = "${length(var.elasticache_subnets)}"

   subnet_id      = "${element(aws_subnet.elasticache.*.id, count.index)}"
-  route_table_id = "${element(aws_route_table.private.*.id, count.index)}"
+  route_table_id = "${var.single_nat_gateway ? aws_route_table.private.id : element(aws_route_table.private.*.id, count.index)}"
 }

 resource "aws_route_table_association" "public" {
```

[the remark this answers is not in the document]
lorengordon: Created https://github.com/gas-buddy/terraform-aws-vpc/pull/1 to try to make this easy.

lorengordon:
@eherot @antonbabenko any more thoughts on this?

antonbabenko:
I'd like to close this PR because it brings unnecessary complexity which should be tested very thoughtfully.

Instead of merging this PR I propose that we always start with a real use-case which is not possible to implement using this module already.

lorengordon:
Oh that's a bummer. The use case is that it is superfluous to create a route table for each private subnet if you have a single nat gateway. Currently, the module will _always_, unnecessarily, create a route table for each private subnet. Maybe it's more a bug than a use case. What do you actually need to see?

PR #3, opened 2017-09-21
Allow the user to define custom tags for route tables

eherot:


antonbabenko:
We now have more tags coverage than tests coverage :trollface: 

Only `aws_db_subnet_group`, `aws_vpc` and `aws_internet_gateway` doesn't have a way to customize tags.

Thanks!

PR #6, opened 2017-09-27
disable dynamodb data source when not needed

mgresko:
dynamodb is not supported in govcloud so this module will always fail to run.  Setting count to false allows this to run correctly.


antonbabenko:
Thanks for the code. Are you sure that DynamoDB is not supported in GovCloud? Could you try it with this module?

According to [AWS documentation](http://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-ddb.html) DynamoDB is actually supported in GovCloud, but I agree to turn it on/off should be optional.

PR #7, opened 2017-09-29
Improve variable descriptions

eheydrick:
No functional changes. Just improvements to the variable description language.


antonbabenko:
Good catch @eheydrick

PR #8, opened 2017-10-02
Fix creation of S3 and DynamoDB VPC Endpoints

alexloginov:


lorengordon:
Curious, what is the rationale for creating a vpc endpoint per subnet?

alexloginov:
@lorengordon Main reason is to avoid NATing S3 and DDB traffic between endpoints. Also, as I understand (please correct me if i'm wrong), default routes will route all subnet traffic thru NET Endpoint if S3/DDB endpoints doens't in subnet.

lorengordon:
Ok, I don't think that happens anyway. The module is already setup to create a route from each private/public route table to the vpc endpoint, so that traffic never traverses the NAT Gateway.

alexloginov:
Ok, it make sense. Anyway, in current version applying template fails if you enable S3 and DDB endpoints, because of broken syntax. This fix it. Do you suggest to create DDB/S3 Endpoint only in 1 private subnet? It is a good idea to route traffic from az1 -> az2 latency-wise?

lorengordon:
> current version applying template fails if you enable S3 and DDB endpoints, because of broken syntax

I've been using this module at v1.0.4 and the master branch, enabling both s3 and ddb endpoints, and it's been working fine. Perhaps open an issue with the error message and a tf file that reproduces the error?

> Do you suggest to create DDB/S3 Endpoint only in 1 private subnet? It is a good idea to route traffic from az1 -> az2 latency-wise?

VPC endpoints are a very interesting networking construct. Their implementation is actually _outside_ the VPC. You'll notice in the endpoint definition that it does not actually associate with any particular subnet. The only manifestation of a VPC endpoint inside the VPC is the endpoint id, which is only a routing target. Therefore, I don't think there is any concern about az->az latency. Or more specifically, the latency concern should be roughly the same from _any_ subnet/az, depending of course on how AWS has implemented this functionality under the covers.

antonbabenko:
Closing this issue. As @lorengordon stated open more specific issue and show the terraform code and error message telling how it is failing.

PR #9, opened 2017-10-04
Use aws_vpn_gateway_route_propagation instead of propagating_vgws

patkar:
Use the `aws_vpn_gateway_route_propagation` resource instead of the `propagating_vgws` property on the `aws_route_table`. This allows to attach additional `aws_vpn_gateway_route_propagation` resources outside of this module and doesn't delete them.

Same principle like using `aws_route` resources instead of the `routes` property within `aws_route_table`.

### Use case

Imagine the following structure to split up states:

```
└── dev
    ├── 10-vpc
    │   └── main.tf
    ├── 20-website
    │   └── main.tf
    └── 30-vpn
        └── main.tf
```

The `10-vpc` uses this module to create the VPC and all subnets. The variables `public_propagating_vgws` and `private_propagating_vgws` are not used, thus both an empty list.

In `30-vpn` a VPN is created and `aws_vpn_gateway_route_propagation` resources are added to the `aws_route_table`s from `10-vpn` (IDs retrieved via remote state).

Problem: If you replan/reapply the VPC module after creating the VPN, the VPC module wants to delete all existing route propagations from `30-vpn` because it defaults to an empty list.
```
Terraform will perform the following actions:

  ~ module.vpc.aws_route_table.public
      propagating_vgws.#:          "1" => "0"
      propagating_vgws.1234567890: "vgw-abcde123" => ""

  ~ module.vpc.aws_route_table.private
      propagating_vgws.#:          "1" => "0"
      propagating_vgws.1234567890: "vgw-abcde123" => ""
```

With using the dedicated `aws_vpn_gateway_route_propagation` resource this is not a problem anymore.

### Additional references

See also the note on [aws_route_table](https://www.terraform.io/docs/providers/aws/r/route_table.html) itself:

> **NOTE on propagating_vgws and the aws_vpn_gateway_route_propagation resource:** If the propagating_vgws argument is present, it's not supported to also define route propagations using aws_vpn_gateway_route_propagation, since this resource will delete any propagating gateways not explicitly listed in propagating_vgws. Omit this argument when defining route propagation using the separate resource.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -28,7 +28,6 @@ resource "aws_route_table" "public" {
   count = "${length(var.public_subnets) > 0 ? 1 : 0}"
 
   vpc_id           = "${aws_vpc.this.id}"
-  propagating_vgws = ["${var.public_propagating_vgws}"]
 
   tags = "${merge(var.tags, var.public_route_table_tags, map("Name", format("%s-public", var.name)))}"
 }
@@ -41,6 +40,13 @@ resource "aws_route" "public_internet_gateway" {
   gateway_id             = "${aws_internet_gateway.this.id}"
 }
 
+resource "aws_vpn_gateway_route_propagation" "public" {
+  count = "${length(var.public_propagating_vgws)}"
+
+  route_table_id = "${aws_route_table.public.id}"
+  vpn_gateway_id = "${element(var.public_propagating_vgws, count.index)}"
+}
+
 #################
 # Private routes
 #################
@@ -48,9 +54,15 @@ resource "aws_route_table" "private" {
   count = "${length(var.azs)}"
 
   vpc_id           = "${aws_vpc.this.id}"
-  propagating_vgws = ["${var.private_propagating_vgws}"]
 
   tags = "${merge(var.tags, var.private_route_table_tags, map("Name", format("%s-private-%s", var.name, element(var.azs, count.index))))}"
+}
+
+resource "aws_vpn_gateway_route_propagation" "private" {
+  count = "${length(var.private_propagating_vgws)}"
+
+  route_table_id = "${aws_route_table.private.id}"
+  vpn_gateway_id = "${element(var.private_propagating_vgws, count.index)}"
 }
 
 ################


main.tf
```diff
@@ -41,18 +40,31 @@ resource "aws_route" "public_internet_gateway" {
...
   vpc_id           = "${aws_vpc.this.id}"
-  propagating_vgws = ["${var.private_propagating_vgws}"]
 
   tags = "${merge(var.tags, var.private_route_table_tags, map("Name", format("%s-private-%s", var.name, element(var.azs, count.index))))}"
 }
 
+resource "aws_vpn_gateway_route_propagation" "private" {
+  count = "${length(var.private_propagating_vgws)}"
+
+  route_table_id = "${aws_route_table.private.id}"
```
antonbabenko: There are multiple private route tables, but one public, so I think this is more correct:
```
count = "${length(var.azs)}"
route_table_id = "${element(aws_route_table.private.*.id, count.index)}"
```

You can test it by setting `private_propagating_vgws = ["123", "456", "789"]`

patkar: Ah, damn it, my fault.

But count on the AZs wouldn't fit too? IMHO every `private_propagating_vgws` entry should be attached to every private route table? What do you think?

patkar: Ah, damn it, my fault.

But count on the AZs wouldn't fit too? This would propagate the first vgw to the first private subnet only and so on, or am I wrong? 

IMHO every `private_propagating_vgws` entry must be attached to every private route table. At least that's the old behaviour. Have to think about, how to do that

antonbabenko:
@patkar Should we close this one?

antonbabenko:
@patkar Could you please take a look at this PR, because it seems pretty much complete. Also we should not rely on number of AZ, but number of private subnets instead (see this - https://github.com/terraform-aws-modules/terraform-aws-vpc/commit/3ddef14f9183bb2afbb39cddb206dfc21e65f7b1).

patkar:
I hope I find time to refactor it the next days :)

robh007:
Hi, 

I've started to try and use this module and came up with a use case were I want to use the VGW that gets created inside the module to enable route propagation, we use direct connect.

Also one thing that is puzzling and I think it's more of an issue with the **route_table** **provider** in **terraform**. Is that within a VPC you can only every attach one **virtual gateway**. So I see no need to specify a list of VGWs. It's currently not possible to add a VGW to the VPC route table that's not attached to the VPC. Also actually passing in the VGW ID into the module previously wouldn't of worked either because the VGW isn't attached to the VPC and the route table or  vpn_gateway_route_propagation resource would then throw an error similar to the one below.

Error: Error applying plan:

1 error(s) occurred:

* aws_vpn_gateway_route_propagation.join: 1 error(s) occurred:

* aws_vpn_gateway_route_propagation.join: error enabling VGW propagation: Gateway.NotAttached: resource vgw-2cedd258

Hopefully no body minds, but I've extended the changes in this pull request to allow you to add VGW route propagation within the module and also if you wish you can create outside and pass in. I'll send a new PR and reference this current PR.

PR #12, opened 2017-10-20
NAT gateway should be tagged too.

michalschott:
NAT gateways are missing tags.


antonbabenko:
Thank you for the contribution!

PR #13, opened 2017-10-24
Add AZ to nat gateway name

felixb:


main.tf
```diff
@@ -139,7 +139,7 @@ resource "aws_nat_gateway" "this" {
   allocation_id = "${element(aws_eip.nat.*.id, (var.single_nat_gateway ? 0 : count.index))}"
   subnet_id     = "${element(aws_subnet.public.*.id, (var.single_nat_gateway ? 0 : count.index))}"
 
-  tags = "${merge(var.tags, map("Name", format("%s", var.name)))}"
+  tags = "${merge(var.tags, map("Name", format("%s-%s", var.name, element(var.azs, count.index))))}"
```
antonbabenko: This will probably give the wrong result when AZs are listed not alphabetically and `single_nat_gateway = true`. Correct me if I am wrong.

felixb: I'm pretty sure it would run fine. But the new change is more explicit and similar to the lines above.

antonbabenko: Yes, this looks better. Thanks.

PR #14, opened 2017-10-25
Fix *dns* vars description according to AWS Docs

qugu:
This is a request to fix description of 2 variables. According to the official documentation, *dns* variables are about public hostnames and the public DNS server(s) of AWS: http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-dns.html

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -46,12 +46,12 @@ variable "azs" {
 }
 
 variable "enable_dns_hostnames" {
-  description = "Should be true if you want to use private DNS within the VPC"
+  description = "Should be true if you want to use public DNS hostnames within the VPC"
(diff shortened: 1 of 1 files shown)


antonbabenko:
Sorry for the delay. I don't know how I managed to skipped this one.

Good idea to write things correctly, but let's write it more generic:
1. `Should be true to enable DNS support in the VPC`
2. `Should be true to enable DNS hostnames in the VPC`

via:
1. https://www.terraform.io/docs/providers/aws/r/vpc.html#enable_dns_support
2. https://www.terraform.io/docs/providers/aws/r/vpc.html#enable_dns_hostnames

Issue #16, opened 2017-10-30
ipv6 support


FernandoMiguel:
Are there plans to include IPv6 support to this module?

antonbabenko:
Yes, but PR is welcome :)

bcenker:
As far as I can tell, there is no simple way to make ipv6 optional due to subnet creation argument requirements and the fact that you do not specify your own ipv6 cidr block.  Here is the behavior I saw:

The ipv6_cidr_block vpc attribute is only created if you enable ipv6 in the aws_vpc definition.  Since you must specify the subnet ipv6_cidr_block argument at the time of subnet creation, and because you must use the ipv6_cidr_block vpc attribute in the calculation of the ipv6 subnet cidr, this vpc attribute MUST exist.  If it does not exist then terraform (as of 0.10.7) will complain when trying to calculate the subnet ipv6_cidr_block block argument.

abbreviated example:
```hcl
resource "aws_vpc" "this" {
  cidr_block           = "${var.cidr}"
  instance_tenancy     = "${var.instance_tenancy}"
  enable_dns_hostnames = "${var.enable_dns_hostnames}"
  enable_dns_support   = "${var.enable_dns_support}"
  --> assign_generated_ipv6_cidr_block = "${var.enable_ipv6}"

  tags = "${merge(var.tags, map("Name", format("%s", var.name)))}"
}
resource "aws_subnet" "public" {
  count = "${length(var.public_subnets)}"

  vpc_id                  = "${aws_vpc.this.id}"
  cidr_block              = "${var.public_subnets[count.index]}"
  --> ipv6_cidr_block         = "${var.enable_ipv6 ? cidrsubnet(aws_vpc.this.ipv6_cidr_block,8,count.index) : ""}"
  availability_zone       = "${element(var.azs, count.index)}"
  map_public_ip_on_launch = "${var.map_public_ip_on_launch}"

  tags = "${merge(var.tags, var.public_subnet_tags, map("Name", format("%s-public-%s", var.name, element(var.azs, count.index))))}"
}
```
This works great if var.enable_ipv6 is set to true.  When set to false, terraform apply will throw this:
```hcl
* module.vpc.aws_subnet.public[0]: Resource 'aws_vpc.this' does not have attribute 'ipv6_cidr_block' for variable 'aws_vpc.this.ipv6_cidr_block'
```

[...]

PR #21, opened 2017-11-05
Issue #16: ipv6 support - add ipv6 support

bcenker:
There are several resources which should be updated to have IPv6 support:
- [x] aws_vpc (done in [v1.39.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.39.0))
- [ ] aws_subnet
- [ ] aws_egress_only_internet_gateway
- [ ] aws_route...

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -7,7 +7,8 @@ resource "aws_vpc" "this" {
   enable_dns_hostnames             = "${var.enable_dns_hostnames}"
   enable_dns_support               = "${var.enable_dns_support}"
   assign_generated_ipv6_cidr_block = "${var.enable_ipv6}"
-  tags                             = "${merge(var.tags, map("Name", format("%s", var.name)))}"
+
+  tags = "${merge(var.tags, map("Name", format("%s", var.name)))}"
 }
 
 ###################
@@ -42,7 +43,7 @@ resource "aws_route" "public_internet_gateway" {
 }
 
 resource "aws_route" "public_internet_gateway_ipv6" {
-  count = "${var.enable_ipv6 ? (length(var.public_subnets) > 0 ? 1 : 0) : 0}"
+  count = "${var.enable_ipv6 && length(var.public_subnets) > 0 ? 1 : 0}"
 
   route_table_id              = "${aws_route_table.public.id}"
   destination_ipv6_cidr_block = "::/0"
@@ -76,7 +77,7 @@ resource "aws_subnet" "public" {
 }
 
 resource "aws_subnet" "public_ipv6" {
-  count = "${ var.enable_ipv6 ? length(var.public_subnets) : 0}"
+  count = "${var.enable_ipv6 ? length(var.public_subnets) : 0}"
 
   vpc_id                          = "${aws_vpc.this.id}"
   cidr_block                      = "${var.public_subnets[count.index]}"
--- a/main.tf
+++ b/main.tf
@@ -2,12 +2,12 @@
 # VPC
 ######
 resource "aws_vpc" "this" {
-  cidr_block           = "${var.cidr}"
-  instance_tenancy     = "${var.instance_tenancy}"
-  enable_dns_hostnames = "${var.enable_dns_hostnames}"
-  enable_dns_support   = "${var.enable_dns_support}"
-
-  tags = "${merge(var.tags, map("Name", format("%s", var.name)))}"
+  cidr_block                       = "${var.cidr}"
+  instance_tenancy                 = "${var.instance_tenancy}"
+  enable_dns_hostnames             = "${var.enable_dns_hostnames}"
+  enable_dns_support               = "${var.enable_dns_support}"
+  assign_generated_ipv6_cidr_block = "${var.enable_ipv6}"
+  tags                             = "${merge(var.tags, map("Name", format("%s", var.name)))}"
 }
 
 ###################
@@ -41,6 +41,14 @@ resource "aws_route" "public_internet_gateway" {
   gateway_id             = "${aws_internet_gateway.this.id}"
 }
 
+resource "aws_route" "public_internet_gateway_ipv6" {
+  count = "${var.enable_ipv6 ? (length(var.public_subnets) > 0 ? 1 : 0) : 0}"
+
+  route_table_id              = "${aws_route_table.public.id}"
+  destination_ipv6_cidr_block = "::/0"
+  gateway_id                  = "${aws_internet_gateway.this.id}"
+}
+
 #################
 # Private routes
 #################
@@ -57,7 +65,7 @@ resource "aws_route_table" "private" {
 # Public subnet
 ################
 resource "aws_subnet" "public" {
-  count = "${length(var.public_subnets)}"
+  count = "${!var.enable_ipv6 ? length(var.public_subnets) : 0}"
 
   vpc_id                  = "${aws_vpc.this.id}"
   cidr_block              = "${var.public_subnets[count.index]}"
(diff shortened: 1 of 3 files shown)


outputs.tf
```diff
@@ -30,6 +30,21 @@ output "private_subnets_cidr_blocks" {
   value       = ["${aws_subnet.private.*.cidr_block}"]
 }
 
+output "ipv6_enabled_private_subnets" {
+  description = "List of IDs of private subnets in an ipv6 enabled VPC"
+  value       = ["${aws_subnet.private_ipv6.*.id}"]
+}
+
+output "ipv6_enabled_private_subnets_ipv4_cidr_blocks" {
+  description = "Map of List of ipv4 cidr_blocks of private subnets in an ipv6 enabled VPC"
```
antonbabenko: The description is a dangerous nice mix of ipv4 and ipv6 even if it makes sense :)

antonbabenko: Also, this is not `map of list`, but just a `list`

main.tf
```diff
@@ -67,11 +75,24 @@ resource "aws_subnet" "public" {
   tags = "${merge(var.tags, var.public_subnet_tags, map("Name", format("%s-public-%s", var.name, element(var.azs, count.index))))}"
 }
 
+resource "aws_subnet" "public_ipv6" {
```
antonbabenko: Making `aws_subnet` resources conditionally is a bad idea, because it require twice more code for some things. Instead, we should always use the same `aws_subnet` resource (`public` and `private`). I think it should be possible, because ipv6-related features are the same as for ipv4 (with the exception of `aws_egress_only_internet_gateway` which is only for ipv6).

Follow up questions:
1. Will `ipv6_cidr_block` work with empty value?
1. What will the value of `aws_vpc.this.ipv6_cidr_block` be when VPC has not enabled IPv6?
1. Why did you specify the number of bits as "8" as an argument of `cidrsubnet`?

bcenker: Hi Anton,
I agree that making ```aws_subnet``` resources conditionally is not ideal - if you had a chance to review the discussion in the related issue you'll see I originally went down the path of using the same ```private``` and ```public``` subnet resources but ran into some issues due to the ```ipv6_cidr_block``` behavior.

To address your questions directly:
1.  ```ipv6_cidr_block``` subnet argument will work with an empty value, but the following 
```ipv6_cidr_block         = "${var.enable_ipv6 ? cidrsubnet(aws_vpc.this.ipv6_cidr_block,8,count.index) : ""}"```
won't work due to the answer to 2. below 
2.  ```aws_vpc.this.ipv6_cidr_block``` is undefined when the ```assign_generated_ipv6_cidr_block``` argument is ```false``` in vpc resource definition, which resulted in the conditional referenced in 1. failing with the following error: 
```* module.vpc.aws_subnet.public[0]: Resource 'aws_vpc.this' does not have attribute 'ipv6_cidr_block' for variable 'aws_vpc.this.ipv6_cidr_block'```
3.  I specified 8 bits in ```cidrsubnet``` because AWS ipv6 blocks are autoassigned a /56 mask and the AWS ipv6 subnets require a /64 mask, so this should always be 8 (i think)

File: main.tf [hunk 3]
```diff
@@ -2,12 +2,12 @@
...
-  instance_tenancy     = "${var.instance_tenancy}"
-  enable_dns_hostnames = "${var.enable_dns_hostnames}"
-  enable_dns_support   = "${var.enable_dns_support}"
-
-  tags = "${merge(var.tags, map("Name", format("%s", var.name)))}"
+  cidr_block                       = "${var.cidr}"
+  instance_tenancy                 = "${var.instance_tenancy}"
+  enable_dns_hostnames             = "${var.enable_dns_hostnames}"
+  enable_dns_support               = "${var.enable_dns_support}"
+  assign_generated_ipv6_cidr_block = "${var.enable_ipv6}"
```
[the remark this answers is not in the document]
bcenker: Fixed in latest commit

main.tf
```diff
@@ -41,6 +41,14 @@ resource "aws_route" "public_internet_gateway" {
   gateway_id             = "${aws_internet_gateway.this.id}"
 }
 
+resource "aws_route" "public_internet_gateway_ipv6" {
+  count = "${var.enable_ipv6 ? (length(var.public_subnets) > 0 ? 1 : 0) : 0}"
```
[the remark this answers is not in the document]
bcenker: oops 🤦‍♂️ :)  Fixed in latest commit

outputs.tf
bcenker: Fixed in latest commit

main.tf
antonbabenko: Thanks for the clarification. I want to try the code in this PR and see if I can make something around 1 and 2 during Monday evening.

bcenker: @antonbabenko looks like there was a recently submitted [PR](https://github.com/terraform-providers/terraform-provider-aws/pull/2103) that would fix the [issue](https://github.com/terraform-providers/terraform-provider-aws/issues/688) we are seeing.

If we wait until this is fixed we won't need to create conditional subnet resources.  Additionally, it looks like there is a "fix" coming in 0.11.x (see 0.11.0-beta1 [changelog](https://github.com/hashicorp/terraform/blob/master/CHANGELOG.md#0110-beta1-november-3-2017)) that may cause this PR code to break because of the new ```output``` validation (I assume that if our conditionally created subnet resources are not created, terraform would throw an error due to the exception in the ipv6_* output interpolations).

I'm thinking we can sit on this until that PR is merged and I can refactor the ipv6 support into the existing the resources, and hopefully at the same time provide forward compatibility.  Let me know your thoughts :)

antonbabenko: Great finding! Absolutely, let's wait for this PR. Terraform AWS provider is the best place to fix this kind of issues.

nelg:
+1 for this feature

nicolas-g:
+1

cornfeedhobo:
Has work on this halted? It's pretty much the last thing in the way of me switching over from my own vpc module ...

antonbabenko:
Please thumb-up the related issue in the AWS provider - https://github.com/terraform-providers/terraform-provider-aws/pull/2103.

[This](https://github.com/terraform-providers/terraform-provider-aws/pulls?q=is%3Apr+is%3Aopen+sort%3Areactions-%2B1-desc) should indicate to the maintainers that this feature is not very large and it is still important for a lot of users.

OneBadSanta:
+1 Still not available in v1.37.0

antonbabenko:
IPv6 is a good feature to have, but it is not so easy to make it supported in this module.

I have updated this module by adding support for`assign_generated_ipv6_cidr_block` to the VPC resource - [v1.39.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.39.0) has been released.

Note, that this adds IPv6 support only to VPC and not to other resources. Adding IPv6 support to `aws_subnet`, `aws_egress_only_internet_gateway` and others requires some time.

rsrchboy:
It looks like the upstream provider PR has been merged in; are we ready to move forward with this PR again?

antonbabenko:
Yes, at least we can try once more. Last time I looked at it there were some bugs which prevented me from making it. I am not entirely sure what was it.

miry:
I suggest split this PR to small parts. Example:

ipv6 for subnets
ipv6 for internet gateways
...

dekimsey:
I re-implemented most of this in #218 with success to test it. I think all the terraform components are in place.

I did notice some issues with the deletion of the ipv6 route table when the vpc is destroyed. 

One last thing, this PR makes an assumption about the spread of the ipv6 subnets. I think it'd be better to have the user specify the prefixes, they already are doing so for ipv4 anyway.

antonbabenko:

PR #29, opened 2017-11-15
make sure outputs are always valid

jbardin:
Outputs directly referencing resource IDs that are only conditionally
created are not valid. Use the concat * trick to make sure we always
have a valid value for these outputs.


antonbabenko:
Thanks for making this repo forward compatible!

lorengordon:
Thanks! We were running into this problem Monday, when compiling terraform from master. Had set it aside until today to troubleshoot, and now it's fixed! 😁 👍

PR #30, opened 2017-11-15
Add tags to the DHCP Options resource

rifelpet:
The aws_vpc_dhcp_options resource [supports tags](https://www.terraform.io/docs/providers/aws/r/vpc_dhcp_options.html#tags) so we should add them to be consistent.


antonbabenko:
Thank you for paying attention to details!

PR #38, opened 2017-11-28
Allow passing in EIPs for the NAT Gateways

### LINKED ISSUE #37: Allow passing in the EIPs for the NAT gateways
The NAT Gateway IPs in my setup need to survive the terraform apply-destroy cycle because they are the interface to integrate my infrastructure with other infrastructures outside my control. More specifically, these IPs need to be whitelisted somewhere else for my system to work.

As it is now, if I build up my infrastructure, destroy it and build it again, I will get new IPs for the NAT Gateways. This poses a problem for me because whitelisting the new IPs has to go through a manual process.

To this end I would like to be able to allocate EIPs outside my plan, and pass those into the VPC module as input variables. I'm opening the issue to raise awareness to the problem and to promote a discussion about it. Maybe there are other ways of achieving this that I'm missing?

Meanwhile I'll work on a PR to address this problem.

miguelaferreira:
This PR adds enables passing in external EIPs to be assigned to the NAT Gateways (Fixes #37).

To that end, this PR:
* Adds two variables: a boolean flag to enable disable the feature and a list of EIP IDs to be used if the feature is enabled.
* Adds the logic to either create new NAT Gateway IPs or use the EIP IDs passed in to the module
* Adds a restriction to the required terraform version (>= 0.10.13) since the selection of either created or passed in IPs requires local values.

I have not documented the feature because I would like some input from the maintainers of the module on where to document this. If the PR is accepted, I could add the documentation to the main README file, to the existing examples, or even make a new example.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,3 +1,7 @@
+terraform {
+  required_version = ">= 0.10.13" # introduction of Local Values configuration language feature
+}
+
 ######
 # VPC
 ######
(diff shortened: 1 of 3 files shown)


main.tf
```diff
@@ -152,16 +156,28 @@ resource "aws_elasticache_subnet_group" "elasticache" {
 ##############
 # NAT Gateway
 ##############
+# Workaround for interpolation not being able to "short-circuit" the evaluation of the conditional branch that doesn't end up being used
```
antonbabenko: Very good explanation.

variables.tf
```diff
@@ -65,6 +65,17 @@ variable "single_nat_gateway" {
   default     = false
 }
 
+variable "reuse_nat_ips" {
+  description = "Should be true if you don't want EIPs to be created for your NAT Gateways and will instead pass them in via the external_nat_ips variable"
```
antonbabenko: Change: external_nat_ips => external_nat_ip_ids

miguelaferreira:
Thank you for taking the time to review my PR. I've pushed the changes you requested. Please let me know if there's anything else.

antonbabenko:
Thanks! [v1.9.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/tag/v1.9.0) is now released.

miguelaferreira:
Great! Thank you for the corrections to the documentation and the fast release of a new version.

PR #42, opened 2017-12-05
Adding additional tags for DHCP option set

shortypowers:
Adding the ability to add custom tags for the DHCP option set


antonbabenko:
[v1.7.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/tag/v1.7.0) is released and it contains this PR. Thanks for the contribution!

Issue #5, opened 2017-09-27
Describe how to update tfstate file after migration to the registry


antonbabenko:
Include instructions from this comment: https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/4#issuecomment-332380572

robinbowes:
I've just run into this (switching to use the new module with live resources that will be destroyed if I run apply).

I found I had to do the following (I'm using terragrunt, but exactly the same commands will work with terraform too):

```
terragrunt state mv module.vpc.aws_internet_gateway.mod module.vpc.aws_internet_gateway.this

terragrunt state mv module.vpc.aws_eip.nateip[0] module.vpc.aws_eip.nat[0]
terragrunt state mv module.vpc.aws_eip.nateip[1] module.vpc.aws_eip.nat[1]
terragrunt state mv module.vpc.aws_eip.nateip[2] module.vpc.aws_eip.nat[2]

terragrunt state mv module.vpc.aws_nat_gateway.natgw[0] module.vpc.aws_nat_gateway.this[0]
terragrunt state mv module.vpc.aws_nat_gateway.natgw[1] module.vpc.aws_nat_gateway.this[1]
terragrunt state mv module.vpc.aws_nat_gateway.natgw[2] module.vpc.aws_nat_gateway.this[2]

terragrunt state mv module.vpc.aws_vpc.mod module.vpc.aws_vpc.this
```

I was then left with the following resources that were destroyed and re-created (which didn't cause me a problem because I am not using them):

```
module.vpc.aws_db_subnet_group.database
module.vpc.aws_elasticache_subnet_group.elasticache
```

The remaining changes were non-destructive (changes to `tags.Name` only).

michalschott:
I'm using flat terraform v.0.10.7, you don't have to provision resource indexes:

```
terraform state mv module.vpc.aws_vpc.mod module.vpc.aws_vpc.this
terraform state mv module.vpc.aws_nat_gateway.natgw module.vpc.aws_nat_gateway.this
terraform state mv module.vpc.aws_internet_gateway.mod module.vpc.aws_internet_gateway.this
terraform state mv module.vpc.aws_eip.nateip module.vpc.aws_eip.nat
```

antonbabenko:
Can you guys submit a PR to describe this process in more details for people who are facing the same problems?

robinbowes:
Ah, I see - you can move all the indexed resources in one command, eg. in my case:
```
[...]

PR #52, opened 2018-01-09
revert resource names to mod instead of this where possible

willejs:
When this repo was migrated, a change was pushed to master without a PR that changed resources names from:
  - "mod" to "this" for almost resources
  - "nateip" to "nat" for the nat eip
  - "natgw" to "this" for the nat gw

These are massive breaking changes when you attempt to move from the community module to the new hosted module. This module will try to recreate nearly all of the resources which for subnets, route tables etc is impossible when you have running infrastructure inside it. One could move all the resources in their terraform state, but when you are using this across many vpcs, in multiple accounts, it is very difficult and risky. 

Therefore, i am proposing the name changes are reverted to maintain backward compatibility for now. This could however cause issues for people who have been using this module since the migration, but there are probably fewer users in this state then existing users using the old module.

In future, if names need to be changed, for example in the case of the nat gw or eip, create before destroy should be added as a lifecycle hook to facilitate zero downtime changes. Either way, a discussion should take place in a PR first so people can assess the impact of these changes.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -5,7 +5,7 @@ terraform {
 ######
 # VPC
 ######
-resource "aws_vpc" "this" {
+resource "aws_vpc" "mod" {
   cidr_block           = "${var.cidr}"
   instance_tenancy     = "${var.instance_tenancy}"
   enable_dns_hostnames = "${var.enable_dns_hostnames}"
@@ -17,7 +17,7 @@ resource "aws_vpc" "this" {
 ###################
 # DHCP Options Set
 ###################
-resource "aws_vpc_dhcp_options" "this" {
+resource "aws_vpc_dhcp_options" "mod" {
   count = "${var.enable_dhcp_options ? 1 : 0}"
 
   domain_name          = "${var.dhcp_options_domain_name}"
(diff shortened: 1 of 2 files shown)


antonbabenko:
The solution to rename things in state file has been described here https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/5 right after release. This is not just a migration of existing community module, but also new features were added to it. You should always stick to the specific version of the module (especially since you can't upgrade to the latest version).

The rename will obviously break things for existing users and does not provide any benefits for them, so I close this one.

dm:
Might be a bit late to bring another view to the table... However, @antonbabenko, I'll leave my thoughts anyway for your consideration:

Prefacing those with the note that these modules are supposedly "HashiCorp Verified Modules", where it's stated "Verified infrastructure modules are reviewed by HashiCorp and actively maintained by contributors to stay up-to-date".

Hopefully these are also mean production-ready, where adequate maintainability and versioning are a definite expectation. Hopefully the intention to rename these has a purpose, and there should be an upgrade path expected for marking as deprecated, the fact that there are no "releases" doesn't really help much. But as @willejs, this migration made the module incompatible, which for those who use it in production mean you're unable to adequately match existing *production* resources, which obviously there was no intention of removing/reinitialising.

I'd point out this is severe enough to warrant a reconsideration on your position above, as new features weren't backwards incompatible to break the transition. And it would be better to perhaps delineate an upgrade path, if you feel these would be the direction to go to in the future.

Regards,
DM

antonbabenko:
I am a bit surprised that this change (referring to "rename" of resource) has **not** been painful for the majority of users who either started from new clean state or upgraded from the community's one, ran `terraform state mv` commands and now keep using this module.

Modules published under [terraform-aws-modules](https://github.com/terraform-aws-modules) organization are having adequate maintainability and versioning strategy. New issues (bugs and features) being reported gets implemented, and a new release is coming out after PR is merged. It surely can be improved, but it has been working like this for pretty long time. I agree that there should be some more information regarding this process, but I have not had time to prioritize it.

stefancocora:
> Modules published under terraform-aws-modules organization are having adequate maintainability and versioning strategy. 
Reality shows that this statement is false because the breaking change added on 2017-Sep-27 moved the semver version from `v1.0.1` to `v1.0.2`.
This versioning change does not signal to users that this is a backwards incompatible breaking change.

The expectation of users using this module which is **in the terraform registry** is that this module is a production-ready, community maintained module with adequate versioning and an upgrade path is expected. (semver)
It would be very useful to the users of this module if the maintainers would create an upgrade path rather than introducing breaking changes and basically telling the users to "sort it out for yourself".

What is the technical argument for not merging this PR and not naming resources back to the expected "mod" ?

willejs:
@antonbabenko I think closing this PR is a good idea. Merging it would actually be contradictory to my original points. You are right, too much time has passed now to merge this as a patch to the original bug without causing more pain for people. The damage is already done in the original commit to master, without bumping semver, or following a PR process. 

Therefore, we are going to have to fork the module and maintain it separately.  I just ask that in future you consider that people are probably using this quite heavily and moving resources around is not a trivial or recommended task, and is a breaking change that should be agreed by others.

PR #35, opened 2017-11-24
Add redshift subnet option

ChairmanTubeAmp:

PR #53, opened 2018-01-10
Support lambda subnet creation

dpetzold:
Hi,

This adds support for creating subnets for lambda. Just let me know if you have questions or comments.

Thank you,


antonbabenko:
I am not sure if this is necessary, because Lambda subnets are no different than other `private_subnets` which can be defined already.

As you see, this module creates `database_subnets` and `elasticache_subnets` as special types of subnets, because it is convenient to create `aws_db_subnet_group` and `aws_elasticache_subnet_group` at the same time. Same for `aws_redshift_subnet_group` (PR #35).

Please correct me if I am missing something what makes Lambda so special here.

dpetzold:
I see. Thank you for looking at this.

PR #57, opened 2018-01-17
Lowercase database subnet group name

memoryleak:
If the name field doesn't contain lower-case only, it throws an error.


antonbabenko:
Good catch. I guess this is the behavior of AWS internals (SDK or API), which we can't influence from Terraform.

antonbabenko:
[v1.15.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/tag/v1.15.0) has been released.

Issue #56, opened 2018-01-13
Tag everything


xocasdashdash:
Hi!

Some resources created by this module are not tagged. Are you open to a PR adding tag support to the rest of the resources like aws_eip for example?

antonbabenko:
Yes, please submit such PRs!

PR #61, opened 2018-01-21
Issue #56: Added tags for elastic ips

xocasdashdash:
Following conversation on #56 I believe the one Elastic IPs was the only resource that wasn't tagged.


antonbabenko:
[v1.16.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/tag/v1.16.0) is released. Thank you!

PR #65, opened 2018-01-25
Add aws provider in module

mhumeSF:
Hoping this can be added to help reduce code when provisioning multiple vpcs.


antonbabenko:
This won't pass. Providers configuration should be specified on the higher level and not in the resource module itself. Here is a [blog post](https://medium.com/@anton.babenko/providers-version-in-terraform-modules-ab35b47a3d14) about it.

mhumeSF:
Ah! Thanks for settin' me straight

PR #66, opened 2018-01-30
Improve documentation about the usage of external NAT gateway IPs

ajardin:
Hello,

First of all, thanks for your work! That module is very useful.

I tried to use External NAT Gateway IPs by following the readme instructions without success until I read the [documentation](https://registry.terraform.io/modules/terraform-aws-modules/vpc/aws/1.9.1?tab=inputs). From my understanding, the variable `external_nat_ip_ids` is not read if the variable `reuse_nat_ips` is not set to `true`. 

I suggest to add a small note about that combination directly into the readme. :wink:

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,13 +76,14 @@ module "vpc" {
 
   enable_nat_gateway  = true
   single_nat_gateway  = false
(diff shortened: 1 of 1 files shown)


antonbabenko:
Thank you!

PR #67, opened 2018-02-03
Extended aws_vpn_gateway use case.

robh007:
This extends using the **aws_vpn_gateway_route_propagation** resource and closes #9 

Previously if you enabled a VPN gateway there was no way to use that gateway within this module. You could also pass in a list of VGWs to associate with a route table type (Public / Private). However this wouldn't have worked either because you need to attach a VGW to a VPC to then allow a route table to turn on route propagation.

Also a VPC can only ever have 1 VGW attached to it. So I'm struggling to understand why a list of VGWs is expected to be passed in when you are only creating 1 VPC.

- added new variable **attach_vpn_gateway**
- added new resource **aws_vpn_gateway_attach**
- modified **aws_vpn_gateway_route_propagation** usage


robh007:
Hi @antonbabenko,

Is this something you can look at / comment on?

Thanks
Rob.

antonbabenko:
Yes, @robh007 , I looked at it already briefly. I will be able to take another look and respond properly in a few hours. Sorry for the delay!

main.tf
```diff
@@ -333,3 +331,24 @@ resource "aws_vpn_gateway" "this" {
...
+
+resource "aws_vpn_gateway_route_propagation" "public" {
+  count = "${var.public_propagating_vgws && var.enable_vpn_gateway || var.public_propagating_vgws && var.attach_vpn_gateway != "default " ? length(var.public_subnets) : 0}"
+
+  route_table_id = "${element(aws_route_table.public.*.id, count.index)}"
+  vpn_gateway_id = "${element(concat(aws_vpn_gateway.this.*.id, aws_vpn_gateway_attachment.vgw.*.vpn_gateway_id), count.index)}"
+}
+
+resource "aws_vpn_gateway_route_propagation" "private" {
+  count = "${var.private_propagating_vgws && var.enable_vpn_gateway || var.private_propagating_vgws && var.attach_vpn_gateway != "default" ? length(var.private_subnets) : 0}"
```
antonbabenko: `count` can be updated like this:

`count = "${var.create_vpc && var.propagate_private_route_tables_vgw && (var.enable_vpn_gateway || var.vpn_gateway_id = "") ? length(var.private_subnets) : 0}"`

variables.tf
```diff
@@ -102,14 +102,19 @@ variable "enable_vpn_gateway" {
   default     = false
 }
 
+variable "attach_vpn_gateway" {
+  description = "ID of VPN Gateway to attach to the VPC"
+  default     = "default"
```
antonbabenko: `default = ""` is better, because it means "it was not specified".

main.tf
```diff
@@ -333,3 +331,24 @@ resource "aws_vpn_gateway" "this" {
...
+
+resource "aws_vpn_gateway_attachment" "vgw" {
+  count = "${var.attach_vpn_gateway != "default" ? 1 : 0}"
+
+  vpc_id         = "${aws_vpc.this.id}"
+  vpn_gateway_id = "${var.attach_vpn_gateway}"
+}
+
+resource "aws_vpn_gateway_route_propagation" "public" {
+  count = "${var.public_propagating_vgws && var.enable_vpn_gateway || var.public_propagating_vgws && var.attach_vpn_gateway != "default " ? length(var.public_subnets) : 0}"
```
antonbabenko: similar here, as for private

robh007:
Hi @antonbabenko,

I've made the changes you've suggested. I've also changed the public  "aws_vpn_gateway_route_propagation" resource from a count of public subnets to 1. This created addtional resources on the same route table.

antonbabenko:
[v1.23.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/tag/v1.23.0) has been released

robh007:
No problem @antonbabenko happy to help.

PR #69, opened 2018-02-06
Manage Default Route Table under Terraform

lawliet89:
So that the default route table will be named and tagged accordingly instead of being unnamed and untagged.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -334,7 +334,6 @@ resource "aws_vpn_gateway" "this" {
   tags = "${merge(var.tags, map("Name", format("%s", var.name)))}"
 }
 
-
 ###########
 # Defaults
 ###########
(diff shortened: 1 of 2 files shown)


lawliet89:
You're right. I'll change the PR to tag `aws_default_route_table` instead. I didn't know that resource existed =X.

lawliet89:
@antonbabenko I've pushed the changes.

antonbabenko:
This change is not as small as it may seem because it reassigns `default` resources. I will test it during tomorrow and extend this PR a bit.

antonbabenko:
[v1.19.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/tag/v1.19.0) has been released.

PR #83, opened 2018-02-19
Creates a single private route table when single_nat_gateway is true

### LINKED ISSUE #82: Create only one private route table when single_nat_gateway is true
Right now, this module is creating a route table for every subnet even when there is a single NAT Gateway. This is unnecessary, as all the route tables are using the same NAT Gateway, and multiple subnets can be associated with the same route table.

I propose that when the user elects to use a single NAT Gateway, all the private subnets should be associated with just a single private route table.

lorengordon:
Fixes #82

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -4,6 +4,7 @@ terraform {
 
 locals {
   max_subnet_length = "${max(length(var.private_subnets), length(var.elasticache_subnets), length(var.database_subnets), length(var.redshift_subnets))}"
+  nat_gateway_count = "${var.single_nat_gateway ? 1 : local.max_subnet_length}"
 }
 
 ######
@@ -80,11 +81,11 @@ resource "aws_route" "public_internet_gateway" {
 # There are so many routing tables as the largest amount of subnets of each type (really?)
 #################
 resource "aws_route_table" "private" {
-  count = "${var.create_vpc && local.max_subnet_length > 0 ? local.max_subnet_length : 0}"
+  count = "${var.create_vpc && local.max_subnet_length > 0 ? local.nat_gateway_count : 0}"
 
   vpc_id = "${aws_vpc.this.id}"
 
-  tags = "${merge(var.tags, var.private_route_table_tags, map("Name", format("%s-private-%s", var.name, element(var.azs, count.index))))}"
+  tags = "${merge(var.tags, var.private_route_table_tags, map("Name", (var.single_nat_gateway ? "${var.name}-private" : format("%s-private-%s", var.name, element(var.azs, count.index)))))}"
 
   lifecycle {
     # When attaching VPN gateways it is common to define aws_vpn_gateway_route_propagation
@@ -203,7 +204,7 @@ locals {
 }
 
 resource "aws_eip" "nat" {
-  count = "${var.create_vpc && (var.enable_nat_gateway && !var.reuse_nat_ips) ? (var.single_nat_gateway ? 1 : length(var.azs)) : 0}"
+  count = "${var.create_vpc && (var.enable_nat_gateway && !var.reuse_nat_ips) ? local.nat_gateway_count : 0}"
 
   vpc = true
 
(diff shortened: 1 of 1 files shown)


main.tf
```diff
@@ -4,6 +4,7 @@ terraform {
 
 locals {
   max_subnet_length = "${max(length(var.private_subnets), length(var.elasticache_subnets), length(var.database_subnets), length(var.redshift_subnets))}"
+  nat_gateway_count = "${var.single_nat_gateway ? 1 : local.max_subnet_length}"
```
antonbabenko: `max_subnet_length` gets longest of all 4 subnet types, while NAT should only be applicable to `private_subnets`

main.tf
```diff
@@ -203,15 +204,15 @@ locals {
 }
 
 resource "aws_eip" "nat" {
-  count = "${var.create_vpc && (var.enable_nat_gateway && !var.reuse_nat_ips) ? (var.single_nat_gateway ? 1 : length(var.azs)) : 0}"
+  count = "${var.create_vpc && (var.enable_nat_gateway && !var.reuse_nat_ips) ? local.nat_gateway_count : 0}"
```
antonbabenko: This should take into account number of `azs`

main.tf
```diff
@@ -296,28 +297,28 @@ resource "aws_route_table_association" "private" {
...
-  route_table_id = "${element(aws_route_table.private.*.id, count.index)}"
+  route_table_id = "${element(aws_route_table.private.*.id, (var.single_nat_gateway ? 0 : count.index))}"
 }
 
 resource "aws_route_table_association" "database" {
   count = "${var.create_vpc && length(var.database_subnets) > 0 ? length(var.database_subnets) : 0}"
 
   subnet_id      = "${element(aws_subnet.database.*.id, count.index)}"
-  route_table_id = "${element(aws_route_table.private.*.id, count.index)}"
+  route_table_id = "${element(aws_route_table.private.*.id, (var.single_nat_gateway ? 0 : count.index))}"
```
antonbabenko: It does not feel right that `aws_route_table_association` related to `database` has anything to do with the value of `var.single_nat_gateway` even if it is right in this particular piece of code.

main.tf
```diff
@@ -203,15 +204,15 @@ locals {
...
+  count = "${var.create_vpc && (var.enable_nat_gateway && !var.reuse_nat_ips) ? local.nat_gateway_count : 0}"
 
   vpc = true
 
   tags = "${merge(var.tags, map("Name", format("%s-%s", var.name, element(var.azs, (var.single_nat_gateway ? 0 : count.index)))))}"
 }
 
 resource "aws_nat_gateway" "this" {
-  count = "${var.create_vpc && var.enable_nat_gateway ? (var.single_nat_gateway ? 1 : length(var.azs)) : 0}"
+  count = "${var.create_vpc && var.enable_nat_gateway ? local.nat_gateway_count : 0}"
```
antonbabenko: Same here

main.tf
lorengordon: I think that logic is incorrect. You can pass in a list of AZs, and pass in fewer private subnets. You end up with more NAT Gateways than would be used.

This "fixes" that logic, so you end up with a number of NAT Gateways that matches `max_subnet_length`, or just one if using `single_nat_gateway`.

Happy to move that to another PR if you like though.

main.tf
lorengordon: From what I can see, that's not how it works currently. All the private-type subnets, including the "extras" (database, redshift, elasticache), are associated to the "private" route tables. This means they all route through the NAT Gateway. The number of NAT Gateways therefore needs to account for all the private-type subnets, which is what `max_subnet_length` does.

main.tf
lorengordon: I debated created a new variable, since yes I am overloading `single_nat_gateway`, but in the end I didn't see it adding value. And this approach simplified some of the logic so I could re-use the local.

Either way I guess. `single_private_route_table`?

lorengordon: I suppose I could also just create a local named `single_private_route_table` and set it to `var.single_nat_gateway`...

lorengordon:
I guess I'd disagree that this is more complicated. In some ways, I think it is _more_ clear than the current code. The use of a well-named local helps make it quite clear to me what is happening. And centralizing that logic in the local also makes it easier to maintain. Have been using and maintaining this code in a fork for six months, periodically merging from upstream/master, and it's been quite solid and easy to follow.

antonbabenko:
It took some time to review and agree with this PR mentally. 👍 

Thank you!

PR #121, opened 2018-04-25
Adds support for ClassicLink DNS

ipmb:


main.tf
```diff
@@ -374,9 +374,10 @@ resource "aws_vpn_gateway_route_propagation" "private" {
 resource "aws_default_vpc" "this" {
   count = "${var.manage_default_vpc ? 1 : 0}"
 
-  enable_dns_support   = "${var.default_vpc_enable_dns_support}"
-  enable_dns_hostnames = "${var.default_vpc_enable_dns_hostnames}"
-  enable_classiclink   = "${var.default_vpc_enable_classiclink}"
+  enable_dns_support             = "${var.default_vpc_enable_dns_support}"
+  enable_dns_hostnames           = "${var.default_vpc_enable_dns_hostnames}"
+  enable_classiclink             = "${var.default_vpc_enable_classiclink}"
+  enable_classiclink_dns_support = "${var.default_vpc_enable_classiclink_dns_support}"
```
antonbabenko: Could you verify if this work with both old and new AWS regions? I remember there was some difference in implementations depending on a region you work with - new regions may not have `classiclinks` concept and it won't work.

A similar difference was between regions which didn't have concepts of IPv6 when they were created and those which has it since the very beginning.

Maybe this is working, but please double check if you can.

ipmb: Ah, I didn't realize this. It's probably not worth the effort considering how old the classiclink stuff is :)

antonbabenko: If this is fixed inside Terraform AWS provider then we can merge this PR. It would be great to have all features of VPC supported by this module.

Could you verify this PR by running code from examples in a couple of regions?

antonbabenko:
EC2 ClassicLink in 2019! 🎉 

[v2.14.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.14.0) has been released.

PR #122, opened 2018-05-04
Provide separate route tables for db/elasticache/redshift

cam-fulcrum:
Hi,

I've created separate route tables for the database, elasticache and redshift subnets.

Currently the db subnets automatically attach to the private subnet (which in turn is NAT'd out). This means the databases by default have a route to the internet.

This breaks the standard 3 tier dmz approach and I believe this shouldn't be the default. If users want to allow this, they can manually add in routes so their database subnets can be NAT'd out.

Let me know what you think and if you require any changes. Thanks for your hard work :smile:

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -98,6 +98,42 @@ resource "aws_route_table" "private" {
   }
 }
 
+#################
+# Database routes
+#################
+resource "aws_route_table" "database" {
+
+  count = "${var.create_vpc && length(var.database_subnets) > 0 ? length(var.database_subnets) : 0}"
+
+  vpc_id = "${aws_vpc.this.id}"
+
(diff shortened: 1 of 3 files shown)


nunofernandes:
+1

antonbabenko:
Thanks for the PR!

Yes, it makes sense to make it slightly differently from how it is implemented now.

Few comments:
1. Make configurable if a user wants to use `private` route tables (as before, by default) OR create individual tables. This will require 3 new boolean variables - `create_database_route_table`, `create_redshift_route_table`, `create_elasticache_route_table` - all set to `false` by default.
1. Update documentation describing your use-case and show piece of code (either in README or make a new example in `examples`).

murarisumit:
Can be get it merged.

antonbabenko:
After long discussions and several PRs I believe this issue is now resolved and can be closed. See https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/132#issuecomment-414071092 for details.

PR #135, opened 2018-05-31
Adding "infra subnets" as a class

madsenibis:
For situations where Lambda functions are used (e.g., with API Gateway) inside the VPC, the Lambdas need to be configured with subnet IDs from which they can allocate ENIs.  The PR simply adds subnets across the AZs for that.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -193,16 +193,16 @@ resource "aws_elasticache_subnet_group" "elasticache" {
 }
 
 #####################################################
-# infra subnets - private subnet with no NAT gateway
+# intra subnets - private subnet with no NAT gateway
 #####################################################
-resource "aws_subnet" "infra" {
-  count = "${var.create_vpc && length(var.infra_subnets) > 0 ? length(var.infra_subnets) : 0}"
+resource "aws_subnet" "intra" {
+  count = "${var.create_vpc && length(var.intra_subnets) > 0 ? length(var.intra_subnets) : 0}"
 
   vpc_id            = "${aws_vpc.this.id}"
-  cidr_block        = "${var.infra_subnets[count.index]}"
+  cidr_block        = "${var.intra_subnets[count.index]}"
   availability_zone = "${element(var.azs, count.index)}"
 
-  tags = "${merge(var.tags, var.infra_subnet_tags, map("Name", format("%s-infra-%s", var.name, element(var.azs, count.index))))}"
+  tags = "${merge(var.tags, var.intra_subnet_tags, map("Name", format("%s-intra-%s", var.name, element(var.azs, count.index))))}"
 }
 
 ##############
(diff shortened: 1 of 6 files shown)


antonbabenko:
Hi Mark,

This is exactly when [private subnet](https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/135/files#diff-7a370d8342e7203b805911c92454f0f4L118) should be used for. Also, private subnets have route table associated.

Let me know if I am missing something here. If not, let's close this PR.

madsenibis:
Happy to not complicate things, but I'm wondering how to configure two classes of private subnets when I configure the VPC.  I want a group of subnets of one size for Kubernetes worker nodes, and these need access to NAT gateway and thereby, to the external internet.  Then I need a set of subnets for Lambda ENI allocation (for API Gateway, basically), and those subnets only need to be able to reach other VPC subnets, not the external internet.  I'd like the two classes of subnets to have different tags so that other systems can query by tag and find all my lambda, or all my k8s worker subnets, across VPCs and regions, etc etc.  You get the picture.  How do I use "private_subnet" twice, basically?

antonbabenko:
Right, thanks for the clarification. This module currently does not support such duality, there is no way to create private subnets with NAT and without NAT in one go. I think we should add this.

The naming (especially when it is +32C in the room) gets very tricky. Mainly, how to name private subnet which does not have NAT. I forgot the scientific term. I think it is called something like `ephemeral private subnet`, but probably not. LAN? Intranet? Lambda is too specific and it is just a name of the service.

When name is there we should duplicate half of the code which takes care of private subnet resources (like routing tables, associations, etc). If you can do this, PR is welcome. If not, let us know and me or someone else may help with this.

Thanks again for opening this issue!

madsenibis:
I agree, lambda is too specific, I can envision placing other backend systems in subnets of this type as well.  I'll work on it a bit, as you say duplicating the half of the code that handles the needed resources, leaving out NAT.  If I need help, I'll ask!

antonbabenko:
Regarding namings... [RFC1918](https://tools.ietf.org/html/rfc1918) says:
```
Hosts within enterprises that use IP can be partitioned into three categories:

      Category 1: hosts that do not require access to hosts in other
                  enterprises or the Internet at large; hosts within
                  this category may use IP addresses that are
                  unambiguous within an enterprise, but may be
                  ambiguous between enterprises.

      Category 2: hosts that need access to a limited set of outside
                  services (e.g., E-mail, FTP, netnews, remote login)
                  which can be handled by mediating gateways (e.g.,
                  application layer gateways). For many hosts in this
                  category an unrestricted external access (provided
                  via IP connectivity) may be unnecessary and even
                  undesirable for privacy/security reasons. Just like
                  hosts within the first category, such hosts may use
                  IP addresses that are unambiguous within an
                  enterprise, but may be ambiguous between
                  enterprises.

      Category 3: hosts that need network layer access outside the
                  enterprise (provided via IP connectivity); hosts in
                  the last category require IP addresses that are
                  globally unambiguous.

   We will refer to the hosts in the first and second categories as
   "private".  We will refer to the hosts in the third category as
   "public".
```

So, there is no official name for Category 1. After some reading and talking to other people in couple chats I'd like to call it `intra_subnets`.

madsenibis:
The name makes sense to me.  I've been using this over the weekend to create some test VPCs for a project, and I end up with private subnets that have route table entries, but no NAT gateway entries, which is what we want for Category 1 RFC1918 networks.  See what you think.

main.tf
```diff
@@ -192,6 +192,19 @@ resource "aws_elasticache_subnet_group" "elasticache" {
   subnet_ids  = ["${aws_subnet.elasticache.*.id}"]
 }
 
+#####################################################
+# infra subnets - private subnet with no NAT gateway
+#####################################################
+resource "aws_subnet" "infra" {
```
antonbabenko: Should be `intra`, not `infra`

madsenibis:
There you go - docs and the example, changed "infra" to "intra", sorry about that - I read it wrong!  Thanks for your help.

antonbabenko:
Merging, great work!

antonbabenko:
https://github.com/terraform-aws-modules/terraform-aws-vpc/commit/9ced5e967f2dfc41293a585dbcb656425ba3238a - I have just run the complete example, and added few important missing resources. In particular intra route table and assign it to intra subnets.

Please let me know if you see something wrong in this commit. Also, [v1.33.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.33.0) has been released.

PR #152, opened 2018-08-08
complete-vpc-w-variables example added

Vadim-Zenin:
complete-vpc with variables.
Configuration in this directory use [complete-vpc] with variables as much as possible.


antonbabenko:
I don't think there is a value in this example and in this PR, because it just adds some values in a very opinionated way.

Terraform documentation already has detailed explanation of how values can be specified (tfvars file, variables, command line argument, env. variables, etc).

PR #159, opened 2018-08-19
Added amazon_side_asn to vpn_gateway

### LINKED ISSUE #149: Need ability to assign ASN to a VGW
The Terraform resource aws_vpn_gateway supports the ability to assign an ASN. In our environment, we need this ability to use a Cisco transit VPC solution.

A variable for input and using it in the [resource](https://www.terraform.io/docs/providers/aws/r/vpn_gateway.html) should resolve it.

antonbabenko:
Closes #149

Unfortunately, in Terraform 0.11 it is not possible to disable argument conditionally by specifying `null`.

This PR will have to wait until Terraform 0.12 is released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ Terraform version 0.10.3 or newer is required for this module to work.
 
 | Name | Description | Type | Default | Required |
 |------|-------------|:----:|:-----:|:-----:|
+| amazon_side_asn | The Autonomous System Number (ASN) for the Amazon side of the gateway. If you don't specify an ASN, the virtual private gateway is created with the default ASN. | string | `` | no |
 | assign_generated_ipv6_cidr_block | Requests an Amazon-provided IPv6 CIDR block with a /56 prefix length for the VPC. You cannot specify the range of IP addresses, or the size of the CIDR block | string | `false` | no |
 | azs | A list of availability zones in the region | string | `<list>` | no |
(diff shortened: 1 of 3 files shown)


Vermyndax:
Thanks @antonbabenko

gpanula:
Maybe use a random number for the amazon side asn if one isn't provided?

```
amazon_side_asn = "${length(var.amazon_side_asn) > 0 ? var.amazon_side_asn : random_integer.random_asn }"
```

https://www.terraform.io/docs/providers/random/r/integer.html

```
resource "random_integer" "random_asn" {
  min = 50000
  max = 64000
 }
```

antonbabenko:
Should be fine now with the default value (see #176)

antonbabenko:
[v1.45.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.45.0) has been released

plektra:
Hmm, this addition was a bit questionable. I happened to notice it when running `terraform plan` on my existing VPC (incl. VPN) configuration using always latest registry module. Changing the ASN from previous random-generated value to a fixed default recreates the whole VPN setup which might not be the desired action.

In this case though it's obvious that the module version should be locked to prevent breaking changes, but though it might be worth mentioning the side effect. It's a bit annoying when there's new variables along a new release of a module which you just need to add to your existing working configuration in order to prevent unwanted changes.

Issue #160, opened 2018-08-28
Adding secondary CIDR Blocks


JohnnyC08:
I've come across an use case where I want to add a secondary CIDR Block and create new subnets with those blocks. I created my VPC with the module so I want to see if I can leverage it to do that.

With `aws_vpc_ipv4_cidr_block_association` we can associate new CIDR blocks, so maybe we can accept a list of `secondary_cidr_blocks` in the module definition and use the length to know how many associations to make.

Now the part I could use some guidance on is how to handle the dependency case. If I'm adding extra CIDR blocks and subnets. I want the vpc created and then the secondary associations created and then everything else created. 

I can take care of this if I get some guidance on how to handle the dependency structure of ensuring the VPC is created and then the associations occur. Maybe we can abstract the vpc resource to an inner module that takes the vpc information and the secondary cidr blocks, then the rest of the main file can reference that module.

Anyways, tell me what you think.

matthiasr:
@JohnnyC08 please do! I also need this and am about to re-do it based on your notes

antonbabenko:
[v1.41.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.41.0) has been released.

PR #161, opened 2018-08-30
Add secondary CIDR block support using a local variable to derive the…

JohnnyC08:
Addresses https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/160

Use a local variable with the `aws_vpc_ipv4_cidr_block_association` resource to assign as there are many secondary CIDR blocks..

To  ensure the secondary blocks get assigned before subnet creation so we can also pass in the subnets utilizing the new CIDR blocks to `private_subnets` and `public_subnets` respectively.

I believe that will work. Please tell me what you think! I like to add tests and I saw some kitchen stuff, if you want me to add some additional tests to that, then I would be more than happy to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,8 @@ Terraform version 0.10.3 or newer is required for this module to work.
 | assign_generated_ipv6_cidr_block | Requests an Amazon-provided IPv6 CIDR block with a /56 prefix length for the VPC. You cannot specify the range of IP addresses, or the size of the CIDR block | string | `false` | no |
 | azs | A list of availability zones in the region | string | `<list>` | no |
 | cidr | The CIDR block for the VPC. Default value is a valid CIDR, but not acceptable by AWS and should be overridden | string | `0.0.0.0/0` | no |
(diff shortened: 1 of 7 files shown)


matthiasr:
Ha dang, just when I got my version of this into shape and went to submit it did I see this 😄 

@JohnnyC08 I also wrote an example, I put it in #162 but since this was all your idea, do you want to just pull that one over?

JohnnyC08:
@matthiasr I just checked your PR and it was beautiful!  I am going to add your example and the output you added! I am so mad I forgot to add that as well 😠 

:) Thanks for helping me make this better.

JohnnyC08:
@matthiasr Thank you so much for the wonderful example and the reminder to add the output

antonbabenko:
Thank you guys for putting this feature together!

It looks solid, and I will give it a try and update some things here and there to be consistent with the rest of the module during next couple hours.

antonbabenko:
Thanks, guys!

[v1.41.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.41.0) has been released.

PR #164, opened 2018-09-07
Enable classiclink everywhere

poolski:
Rather than having the option to enable ClassicLink only for the `default` VPC, this PR allows users to enable ClassicLink for any VPCs they create. 

This is especially useful if you're looking to use Terraform to migrate from an EC2 Classic environment over to VPC but also want to create and manage your own VPCs rather than sticking with the default that AWS provides.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ To that end, it is possible to assign existing IPs to the NAT Gateways.
 This prevents the destruction of the VPC from releasing those IPs, while making it possible that a re-created VPC uses the same IPs.
 
 To achieve this, allocate the IPs outside the VPC module declaration.
+
 ```hcl
 resource "aws_eip" "nat" {
   count = 3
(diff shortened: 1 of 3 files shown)


antonbabenko:
Thanks @poolski !

Looks good in general.

Could you try to run the example code of this module (`complete-vpc`) in AWS account which was created since 2014 (or so) AND in regions where `classiclink` has never existed as a concept. I think it will fail because AWS returns different structure of the response (see comments in #121 for related change). 

If it works, let's get this merged.

poolski:
You're right in that it will fail in regions where ClassicLink has never existed in the first place, but when run in regions where ClassicLink **does** exist (for legacy accounts), the API call passes:
```
terraform apply  --target=module.platform.module.vpc.aws_vpc.this

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```
But nothing actually happens in AWS. 

Given that it's `false` by default, there's little chance of someone triggering it by accident, but it might help folks with legacy EC2 infrastructure.

PR #175, opened 2018-10-01
Fix/amazon side asn - fix for #149

tbugfinder:
fix for #149 

https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/149


tbugfinder:
maybe should build on top of the PR

PR #176, opened 2018-10-01
change to Amazon default ASN (as per API)

tbugfinder:
https://docs.aws.amazon.com/cli/latest/reference/ec2/create-vpn-gateway.html


antonbabenko:
Thanks @tbugfinder !

PR #179, opened 2018-10-15
Create input "create_database_internet_gateway_route"

lukvdborne:
This will allow easy creation of publicly accessible RDS endpoints instead of having to add a separate `aws_route` resource next to/outside of the `module "vpc { .. }"` block.

Default value is false for full backward compatibility.


antonbabenko:
[v1.49.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/tag/v1.49.0) has been released with this feature.

PR #212, opened 2019-02-11
Add support old db_subnet_group naming (before the move to the registry)

schmidp:
Hi,

not sure if something like this is wanted, but I transitioned to this repo and I added this compatibility flag so the database security group naming stays the same.

All the best, Philipp


antonbabenko:
Thanks for the PR, but I don't think there is a need to have this module so much compatible to the one which was there before.

Things in this module have moved forward very much. This module is more than a year old already.

PR #213, opened 2019-02-12
Added option to create ECR api and dkr endpoints

michieldhadamus:
Not sure if it is better to keep this out of the module but useful for current project I'm working on.


antonbabenko:
Thanks Michiel!

[v1.55.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.55.0) has been released.

PR #214, opened 2019-02-12
Add lambda subnet

cwcoffman:
Updates to main.tf, variables.tf, outputs.tf, and README.md to add optional Lambda subnets, per emerging best practices: https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html#lambda-vpc

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,8 @@ By default, the module will determine the number of NAT Gateways to create based
 ```hcl
 database_subnets    = ["10.0.21.0/24", "10.0.22.0/24"]
 elasticache_subnets = ["10.0.31.0/24", "10.0.32.0/24"]
-private_subnets     = ["10.0.1.0/24", "10.0.2.0/24", "10.0.3.0/24", "10.0.4.0/24", "10.0.5.0/24"]
+private_subnets     = ["10.0.1.0/24", "10.0.2.0/24", "10.0.3.0/24" , "10.0.4.0/24", "10.0.5.0/24"]
+lambda_subnets      = ["10.0.1.0/24", "10.0.2.0/24", "10.0.3.0/24"]
 redshift_subnets    = ["10.0.41.0/24", "10.0.42.0/24"]
 intra_subnets       = ["10.0.51.0/24", "10.0.52.0/24", "10.0.53.0/24"]
 ```
@@ -130,9 +131,9 @@ If `one_nat_gateway_per_az = true` and `single_nat_gateway = false`, then the mo
 
 By default, if NAT Gateways are enabled, private subnets will be configured with routes for Internet traffic that point at the NAT Gateways configured by use of the above options.
 
-If you need private subnets that should have no Internet routing (in the sense of [RFC1918 Category 1 subnets](https://tools.ietf.org/html/rfc1918)), `intra_subnets` should be specified. An example use case is configuration of AWS Lambda functions within a VPC, where AWS Lambda functions only need to pass traffic to internal resources or VPC endpoints for AWS services.
+If you need private subnets that should have no Internet routing (in the sense of [RFC1918 Category 1 subnets](https://tools.ietf.org/html/rfc1918)), `intra_subnets` should be specified. An example use case is configuration of AWS Lambda functions within a VPC, where AWS Lambda functions only need to pass traffic to internal resources or VPC endpoints for AWS services. 
  
-Since AWS Lambda functions allocate Elastic Network Interfaces in proportion to the traffic received ([read more](https://docs.aws.amazon.com/lambda/latest/dg/vpc.html)), it can be useful to allocate a large private subnet for such allocations, while keeping the traffic they generate entirely internal to the VPC.
+Since AWS Lambda functions allocate Elastic Network Interfaces in proportion to the traffic received ([read more](https://docs.aws.amazon.com/lambda/latest/dg/vpc.html)), it can be useful to allocate a large private subnet for such allocations, while keeping the traffic they generate entirely internal to the VPC.  For this purpose, we have also added the optional `lambda_subnets` option per best practices ([read more](https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html#lambda-vpc)).  
 
 You can add additional tags with `intra_subnet_tags` as with other subnet types.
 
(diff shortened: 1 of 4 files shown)


antonbabenko:
Hi Chad!

I am not sure I understand the difference between new `lambda_subnets` and existing `intra_subnets`. If necessary `intra_subnets` can be as large as you want. Could you double check, please.

cwcoffman:
Anton;

The intra subnets I can see are only for internal (network) use - no access to the public.  The Lamba subnet is basically a mirror of the 'private' subnet, but only for use by Lambdas.  Put differently, it's for the use case where Lambda functions need access to the internet.  This is one of the most commonly recommended patterns by AWS for Lambdas in a VPC.

I can see the argument, that you might want some lambdas to be private as well, and I certainly sympathize with wanting to keep this module clean.  Our challenge is that we are finding that, based on best practice, a) we need a truly separate subnet for only lambdas (address space for ENIs), and b) we are also finding that internet access to the internet through a NAT Gateway is usually needed.  The reason we landed on this for all our Lamba functions is that it's pretty rock solid security if they are only accessing the internet through the NAT Gateway - so why not do it all the time.

Look forward to your feedback and .. BTW thanks so much for doing this!!!

FernandoMiguel:
A private subnet, is a private subnet... No matter if it is for Web servers, or for RDS subnet groups.

So, create just another subnet and use it for lambda 


FYI do keep in mind very soon aws is changing the away ENI are attached to lambda, reducing the start time

cwcoffman:
You could make the same argument for “database_subnets”.  Why make a separate subnet for DB?

It is still best practice to create a subnet dedicated to Lambas, per AWS recommendation. From: https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html
•  Create dedicated Lambda subnets in your VPC:

  *   This will make it easier to apply a custom route table for NAT Gateway traffic without changing your other private/public subnets. For more information, see Configuring a Lambda Function to Access Resources in an Amazon VPC<https://docs.aws.amazon.com/lambda/latest/dg/vpc.html>
  *   This also allows you to dedicate an address space to Lambda without sharing it with other resources.

antonbabenko:
`database_subnets` exists because there is a way to create database subnet group and database routing table if necessary.

I agree with what @FernandoMiguel is saying and with the fact that AWS is working on changing the slow part related to ENI and cold-starts.

I think it is better to keep this module smaller and I don't think we need to add support for dedicated `lambda_subnets`.

cwcoffman:
Thank you, gentlemen, I appreciate the look.

Just for clarity, however, my concerns are not slowness related to cold starts.  It’s strictly around scaling up/down, and reserving the IP space.  In addition, as I indicated, this is a best practice.  I’ll plan to keep our fork internal, and continue to maintain internally as you make updates.  I do hope you’ll consider this best practice (again, from AWS, not me) as you move forward with additional design.

antonbabenko:
Can you use `private_subnets` and allocate `/17` CIDR blocks? `/17` gives 32K of IPs. Sounds like a lot to me.

FernandoMiguel:
I've not kept up with this module in depth.
But years ago I forked it to have multiple private subnets. 

I remember seeing last year a PR to add support for N number of subnet.

Which brings to my original suggestion :
Can't you just create more private subnets and label them lambda? 
(assuming the PR I mentioned has been

cwcoffman:
I totally love that idea, Fernando!  I’m not sure how I missed that, but it looks like this is a possibility.  I’m not sure, however, that this was implemented, but this would be far preferable so that it is extensible.

My understanding was that this:

[cid:image001.png@01D4C37B.8546CB30]


Requires an implementation like this – put differently, this is for listing the CIDR ranges, rather than listing the subnet names to be created.

[cid:image002.png@01D4C37B.8546CB30]

This was why I created the new, named subnet.  Looking forward to your feedback!

FernandoMiguel:
sorry @cwcoffman but you need to post to github... email will mangle all your replies and not included attachemnts

cwcoffman:
Thanks for the feedback @FernandoMiguel - I cleaned up the posts and will only use Github in the future.

FernandoMiguel:
technically i don't oppose having named subnets :)
just dont see the need for one called lambda
if this PR allows for adding N private subnets, I would use it...
heck, i'm forking this for that sole reason

cwcoffman:
@FernandoMiguel - I actually like your approach better, as named subnets are probably OK, but having 'types' of subnets that can be named and extended for whatever purpose is definitely better for the community.  When your fork is in, I'd love to see it!

FernandoMiguel:
eh... i never got to clean it up and push it 
ROFL
i actually need to do the opposite... pull master to mine.. i'm quite far behind

but as far as i remember, the only reason was pretty much just needing 4 private named subnets

it doesnt scale ... it's just hardcoded 4 sets of private subnets.

so feel free to improve yours to included N and label them :D

antonbabenko:
I think it would be a good idea to support the tagging of individual resources (like subnets) by index, like this?
```
# terraform.tfvars
private_subnet_tags_index = {
  0 = {"Name" = "private-for-lambda", "purpose" = "lambda"}
  1 = {"Name" = "private-for-everything", "purpose" = "everything"}
}
```

Would this be sufficient for discovering subnets using data source?

cwcoffman:
Anton - love this idea. If we had n number of private, intra, public subnets and the ability to tag like this, I think that would be pretty ideal.  Then as services / best practices change, we can respond with new and improved infrastructure fairly rapidly.

FernandoMiguel:
my only problem with that data structure in TF <0.12 is that if you want to delete one of the index, it will destroy and recreate all the following resources, right?

a bit dangerous

antonbabenko:
@FernandoMiguel It won't delete/recreate resources because tags can be changed without downtime and `private_subnet_tags_index` is a `map`, not `list`, so it won't be recreated if middle-element changes.

syalam:
> I think it would be a good idea to support the tagging of individual resources (like subnets) by index, like this?
> 
> ```
> # terraform.tfvars
> private_subnet_tags_index = {
>   0 = {"Name" = "private-for-lambda", "purpose" = "lambda"}
>   1 = {"Name" = "private-for-everything", "purpose" = "everything"}
> }
> ```
> Would this be sufficient for discovering subnets using data source?

+1 - this is an elegant way of keeping private subnets agnostic of different workloads/requirements

antonbabenko:
Closing this PR.

PR #217, opened 2019-02-14
New shared subnet 

ppieprzycki:
Hello All,
I would like to propose a new type of network and route table called shared networks. 
In our organization, we need new subnet type that we can then share to other AWS accounts via technologies like vpc peering and vpc share. Inside our network we establish vpc peering between account and only shared route table has routing information regarding cidrs of other vpcs in different accounts. 
We don't want to expose our private or intra networks to other accounts within AWS Organization. 

Please find diagram and more info in the example directory
https://github.com/ppieprzycki/terraform-aws-vpc/tree/new_shared_subnet/examples/vpc-shared-subnets-and-rt


antonbabenko:
Hi Piotr,

I can see what you are trying to achieve, but I don't think this is so generic use-case to be accepted as part of this module.

Normally, people use VPC peering between private networks, and if this is not possible I recommend to implement additional resources next to calling this module.

Closing this PR.

PR #218, opened 2019-02-14
Adds ipv6 support with user-specified prefixes

dekimsey:
Based on work done in #21. There didn't seem to be much going on that pull request. I apologize if this is undesirable.

Adds support for what I think are all the necessary components.

- [x] vpc
- [x] routes
- [x] gateways (egress only for private & database, internet gateway for public)
- [x] prefix assignment
- [x] infra subnet
- [ ] special subnets (elasticache, redshift, etc)
- [x] default vpc security group to permit egress traffic
- [ ] Currently two flags (`enable_ipv6`, `assign_generated_ipv6_cidr_block`), but I don't think that's good. Since they both must be set, and it doesn't seem to make sense to enable one without the other.
- [ ] Destroying the VPC appears to be unclean, results in `* aws_route...._ipv6_egress.2: InvalidRoute.NotFound: no route with destination-cidr-block ::/0 in route table rtb-0be2b413b6e783fc7` error messages. Not sure why, ipv4 route entries clean up correctly. Bug in provider?

I took a slightly different approach particularly on the prefix assignments as I found changing the assignments can cause an [error](https://github.com/terraform-providers/terraform-provider-aws/issues/1519) so I didn't want to use any `+length()` or hard-coded subnet spacing.

Usage:

```
variable "vpc_netblocks" {
  # third octet ~ availability zone
  description = "Network subnet layout"

  default = {
    vpc                = ["10.1.0.0/16"]
    product-private    = ["10.1.12.0/24", "10.1.14.0/24", "10.1.16.0/24"]
    product-private-ipv6-prefixes = [12, 14, 16]
   }
}

module "vpc" {
  ...
  private_subnets = "${var.vpc_netblocks["product-private"]}"
  private_subnet_ipv6_prefixes  = "${var.vpc_netblocks["product-private-ipv6-prefixes"]}"
  assign_generated_ipv6_cidr_block = true
  assign_ipv6_address_on_creation = true
  enable_ipv6                      = true
}
```

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -315,6 +315,9 @@ resource "aws_subnet" "intra" {
   cidr_block        = "${var.intra_subnets[count.index]}"
   availability_zone = "${element(var.azs, count.index)}"
 
+  # TODO: Does not unset, need null support or the like see: Terraform 0.12
+  ipv6_cidr_block = "${var.enable_ipv6 && length(var.intra_subnet_ipv6_prefixes) > 0 ? cidrsubnet(aws_vpc.this.ipv6_cidr_block, 8, element(concat(var.intra_subnet_ipv6_prefixes, list("0")), count.index)) : ""}"
+  
   tags = "${merge(map("Name", format("%s-intra-%s", var.name, element(var.azs, count.index))), var.tags, var.intra_subnet_tags)}"
 }
 
--- a/main.tf
+++ b/main.tf
@@ -70,7 +70,7 @@ resource "aws_internet_gateway" "this" {
 }
 
 resource "aws_egress_only_internet_gateway" "this" {
-  count = "${var.enable_ipv6 && length(var.database_subnets) > 0 && length(var.private_subnets) > 0 ? 1 : 0}"
+  count = "${var.enable_ipv6 && length(var.database_subnets) + length(var.private_subnets) > 0 ? 1 : 0}"
 
   vpc_id = "${aws_vpc.this.id}"
 }
--- a/main.tf
+++ b/main.tf
@@ -33,6 +33,18 @@ resource "aws_vpc_ipv4_cidr_block_association" "this" {
   cidr_block = "${element(var.secondary_cidr_blocks, count.index)}"
 }
 
+resource "aws_security_group_rule" "ipv6-egress" {
+  # VPC default has egress out
+  count            = "${var.create_vpc && var.enable_ipv6 ? 1 : 0}"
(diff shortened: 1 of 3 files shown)


dekimsey:
~~Remaining issue: In my VPC, my box gets an ipv6 address but seems unable to route. I might have done something incorrect with my routing rules but I'm not seeing up. I hope someone can spot my error here.~~ 

Leaving this for posterity. I initially reported IPv6 did not work for me though I felt I had implemented the module correctly. I dug into it today and it appears to be an issue with CentOS (7.6?) that Router Advertisements are not enabled. The default RA-issue route was not set. Enabling the sysctl setting `net.ipv6.conf.eth0.accept_ra=1` immediately triggered getting correct routes. See: https://bugzilla.redhat.com/show_bug.cgi?id=1498854 and https://access.redhat.com/solutions/347693#comment-520173

main.tf
```diff
@@ -69,6 +69,12 @@ resource "aws_internet_gateway" "this" {
   tags = "${merge(map("Name", format("%s", var.name)), var.tags, var.igw_tags)}"
 }
 
+resource "aws_egress_only_internet_gateway" "this" {
+  count = "${var.enable_ipv6 && length(var.database_subnets) > 0 && length(var.private_subnets) > 0 ? 1 : 0}"
```
StephenKing:
```suggestion
  count = "${var.enable_ipv6 && length(var.database_subnets) + length(var.private_subnets) > 0 ? 1 : 0}"
```
Otherwise this is false with `length(var.database_subnets) == 0`

variables.tf
```diff
@@ -18,6 +18,40 @@ variable "assign_generated_ipv6_cidr_block" {
   default     = false
```
StephenKing: would it make sense to remove this variable and set this `assign_generated_ipv6_cidr_block = "${var.enable_ipv6}"`?

StephenKing: Ah.. now I see that you actually mentioned this in the PR description..

dekimsey: Yup, I didn't want to remove the option until I had a chance to hear how you wanted to handle it. I wanted to remove the `assign_generated_ipv6_cidr_block` variable entirely and just set have the new variable name be `enable_ipv6`. Alternatively, keeping the assign name and making it actually set up the ipv6 cidr associations. Do you have any thoughts on how you'd like it to be handled?

antonbabenko:

PR #168, opened 2018-09-13
add vars for custom subnet and route table names

RupertExact:

PR #220, opened 2019-02-18
Added intra subnet suffix.

### LINKED ISSUE #219: Custom suffix for intra subnets
Looks like intra subnets were left out of #168 . I'll make a PR here shortly.

CC: @RupertExact @antonbabenko

kmcquade:
Fixes #219

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -174,7 +174,7 @@ resource "aws_route_table" "intra" {
 
   vpc_id = "${local.vpc_id}"
 
-  tags = "${merge(var.tags, var.intra_route_table_tags, map("Name", "${var.name}-intra-${var.intra_subnet_suffix}"))}"
+  tags = "${merge(var.tags, var.intra_route_table_tags, map("Name", "${var.name}-${var.intra_subnet_suffix}"))}"
 }
 
 ################
--- a/main.tf
+++ b/main.tf
@@ -174,7 +174,7 @@ resource "aws_route_table" "intra" {
 
   vpc_id = "${local.vpc_id}"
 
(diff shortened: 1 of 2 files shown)


main.tf
```diff
@@ -174,7 +174,7 @@ resource "aws_route_table" "intra" {
 
   vpc_id = "${local.vpc_id}"
 
-  tags = "${merge(map("Name", "${var.name}-intra"), var.tags, var.intra_route_table_tags)}"
+  tags = "${merge(var.tags, var.intra_route_table_tags, map("Name", "${var.name}-intra-${var.intra_subnet_suffix}"))}"
```
antonbabenko: The order should be as before so that `Name` tag can be overridden.

Also, `intra` is duplicated in value.

kmcquade: @antonbabenko - see the updated commit. I removed the duplicate intra. 
Not sure what you mean by "The order should be as before so that `Name` tag can be overridden." It's the same format as the previous fixes (see line 166, one of the mods introduced from #168 ), and includes the Name tag which will overwrite anyway.

antonbabenko: Line 177 should have `map(), tags, intra_subnet_tags` - in this order. This is for the reason that more specific tags can override upstream. (`intra_subnet_tags => tags => map()`)

See line 284 for the correct example.

kmcquade: @antonbabenko - thanks for clarifying. Updated.

antonbabenko:
[v1.56.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.56.0) has been released.

PR #222, opened 2019-02-25
Redshift public subnets

bmihaescu:
I updated the VPC module by adding a public subnet for Redshift because in our use case we needed to access it from Kinesis. 
Because VPC endpoint for [Kinesis is not yet supported by Redshift](https://docs.aws.amazon.com/redshift/latest/mgmt/enhanced-vpc-working-with-endpoints.html) , we had to add Redshift to a public subnet and made it publicly accessible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ Sometimes it is handy to have public access to RDS instances (it is not recommen
 
 ## Public access to Redshift cluster
 
-Sometimes it is handy to have public access to Redshift clusters (for example if you need to access it by kinesis - VPC endpoint for kinesis is not yet supported by redshift) by specifying these arguments:
+Sometimes it is handy to have public access to Redshift clusters (for example if you need to access it by Kinesis - VPC endpoint for Kinesis is not yet supported by Redshift) by specifying these arguments:
 
 ```hcl
   create_redshift_public_subnet_group       = true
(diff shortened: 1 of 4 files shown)


bmihaescu:
@antonbabenko can you review this please ?

Thanks

antonbabenko:
@bmihaescu Thanks for this PR.

I've just looked into it once more and I think you have overengineered it a bit. Let me explain.

Users should be able to create redshift subnets as before, but they also should have a possibility to control one thing - `route_table_id` in `resource "aws_route_table_association" "redshift"` depending on whether they want it to be via public or private:
```
route_table_id = "${element(coalescelist(aws_route_table.redshift.*.id, aws_route_table.private.*.id), (var.single_nat_gateway || var.create_redshift_subnet_route_table ? 0 : count.index))}"
```

This means when user wants to change property and make it public or private just another routing table will have to be reassociated.

Remove creation of `redshift_public` resources.

Also, please make sure that this change is backward compatible by setting correct default values.

bmihaescu:
@antonbabenko I modified the code with your suggestion. 
I also tested backward compatibility and all seems good. 
Can you review it once again please ? 

Thanks

antonbabenko:
Thanks @bmihaescu !

This looks pretty good. I will merge it tomorrow.

antonbabenko:
[v1.59.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.59.0) has been released.

PR #229, opened 2019-03-11
documents migration from previous community module 

### LINKED ISSUE #5: Describe how to update tfstate file after migration to the registry

christianuhlcc:
Hi, since I got confused upgrading and had to dig around in the issues (#4) I thought it would be nice to document it in the Readme

 fixes #5

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,6 +375,20 @@ gem install bundler; bundle install
 ```
 3. Test using `bundle exec kitchen test` from the root of the repo.
 
+## Migration
+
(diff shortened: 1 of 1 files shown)


antonbabenko:
Hi @christianuhlcc !

I am not sure we have to keep this information in such a detailed way in README. This module was released 1,5 years ago and moved very far ahead.

I agree that we may have a note about a _possibility_ to migrate to the initial version of this module which was released 1,5 years ago, and link to the related issue for the reference (as a starting point in the migration process). Some commands may be different depending on what resources he had created earlier.

Will you be able to rewrite the text?

christianuhlcc:
I wrote the doc in a way that it would have helped me a lot when I needed to migrate a few days ago. If you think that is not relevant (or everybody except me already migrated) we can just close the PR, because the issues are available to dig through.

PR #239, opened 2019-03-25
Update outputs.tf

vikashealthtap:


vikashealthtap:
i want count of private subnets in output section

antonbabenko:
Not sure what this code is supposed to be used for, but you can use `${length(module.vpc.private_subnets)}` to get number of subnets outside of this module.

vikashealthtap:
i have to create a resource in my vpc file ?

Issue #207, opened 2019-01-30
Feature Requeest - AWS Transit Gateway Support


Jeeppler:
AWS announced a new product during re:Invent 2018 called [Transit Gateway](https://aws.amazon.com/transit-gateway/). Transit Gateway (TGW) acts as a hub between connected VPC's and supports VPN as well.

The VPC reference architecture from AWS suggests to have separate subnets just for the TGW endpoints [1]. 

![tgw_vpc_architecture](https://user-images.githubusercontent.com/2007739/52001734-c53efa80-2485-11e9-81e4-b0e0c28ea96c.png)

It would be nice if this module could support the creation of separate TGW subnets and have in general TGW support.

[1]: AWS re:Invent 2018: [NEW LAUNCH!] Introducing AWS Transit Gateway (NET331)  https://youtu.be/yQGxPEGt_-w?t=24m51s

antonbabenko:
Yes, this would be a great addition to this module. PR is welcome.

Jeeppler:
@antonbabenko I am not able to create a PR for the Transit Gateway feature.  However, I would be really happy if somebody else wants to take on the task.

antonbabenko:
You should fork this repository, add required code and submit a pull-request into this repository.

tbugfinder:
IMHO, terraform-aws-providers still lacks a few TGW details, however we could start with an implementation to start with.

tbugfinder:
Do you think the TGW should be part of the terraform-aws-vpc module as it is with the "VPC service" ?

antonbabenko:
TBH I have not been thinking about this feature so much yet, but I think it makes sense to be part of this module in the first place. This module can handle the creation of TGW, routes, route tables, VPC attachments.

Also, it should be possible to attach other resources to it outside of the module, if users want.

Jeeppler:
At the very least this module should be able to create subnets for the TGW endpoins.

alex-goncharov:
Personally, I do not think that TGW itself has a place in this module, given that multiple VPC's share one or more TGW's. Then, there are TGW routing tables, which, i think, belong to TGW and are also "shared" among VPC's. It is also possible that TGW and routing tables are in different account.

What could be a part of this module are VPC attachments to the TGW as well as additional routes that point to a given TWG. I looked at it and it seems (at least until tf 12) that using route table IDs to populate additional routes elsewhere is easier than passing routing information into module.

mehstg:
[...]

PR #241, opened 2019-03-29
Feature/TransitGateWay

ebalakumar:
TLDR:
The PR has changes for creating a Transit Gateway , attach with the created VPC and then add a route to the TGW in VPC routing tables . 

In Depth : 
1. Creates a TGW based on boolean variable . 
2. Creates TGW Attachment for the VPC . Based on variable , private or public subnets are selected.
3. Creates a TGW Route table , adds the attachment and creates a propagation .
4. AWS_ROUTEs is added to the route tables . CIDR value is received as input and its targeted to TGW.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -939,11 +939,11 @@ resource "aws_ec2_transit_gateway_vpc_attachment" "this" {
 }
 
 resource "aws_ec2_transit_gateway_route_table" "this" {
-  count  = "${(var.create_tgw || var.attach_tgw) && (var.subnet_type_tgw_attachment == "private" || var.subnet_type_tgw_attachment == "public")? 1 : 0}"
+  count = "${(var.create_tgw || var.attach_tgw) && (var.subnet_type_tgw_attachment == "private" || var.subnet_type_tgw_attachment == "public")? 1 : 0}"
 
-  transit_gateway_id = "${aws_ec2_transit_gateway.this.id}"
-  tags               = "${merge(map("Name", format("%s", var.name)), var.tags, var.tgw_tags)}"
+  transit_gateway_id = "${element(concat(aws_ec2_transit_gateway.this.*.id, list(var.attach_tgw_id)), 0)}"
 
+  tags = "${merge(map("Name", format("%s", var.name)), var.tags, var.tgw_tags)}"
 }
 
 #########
--- a/main.tf
+++ b/main.tf
@@ -906,3 +906,76 @@ resource "aws_default_vpc" "this" {
 
   tags = "${merge(map("Name", format("%s", var.default_vpc_name)), var.tags, var.default_vpc_tags)}"
 }
+
+#########
+# Transit Gateway
+#########
+
+resource "aws_ec2_transit_gateway" "this" {
+  count                           = "${var.create_tgw && (var.subnet_type_tgw_attachment == "private" || var.subnet_type_tgw_attachment == "public")? 1 : 0}"
+  description                     = "Transit Gateway"
+  default_route_table_association = "disable"
+  default_route_table_propagation = "disable"
+  tags                            = "${merge(map("Name", format("%s", var.name)), var.tags, var.tgw_tags)}"
+}
+
+# #########
+# # Transit Gateway Attachment
+# #########
+
+resource "aws_ec2_transit_gateway_vpc_attachment" "this" {
+  count                                           = "${var.create_tgw && (var.subnet_type_tgw_attachment == "private" || var.subnet_type_tgw_attachment == "public") && length(var.public_subnets) > 0 ? 1 : 0}"
+  subnet_ids                                      = ["${split(",", var.subnet_type_tgw_attachment == "private" ? join(",", aws_subnet.private.*.id ) : join(",", aws_subnet.public.*.id))}"]
+  transit_gateway_id                              = "${aws_ec2_transit_gateway.this.id}"
+  vpc_id                                          = "${aws_vpc.this.id}"
+  transit_gateway_default_route_table_association = false
+  transit_gateway_default_route_table_propagation = false
+  tags                                            = "${merge(map("Name", format("%s", var.name)), var.tags, var.tgw_tags)}"
+  depends_on                                      = ["aws_ec2_transit_gateway.this"]
+}
+
+resource "aws_ec2_transit_gateway_route_table" "this" {
+  count              = "${var.create_tgw && (var.subnet_type_tgw_attachment == "private" || var.subnet_type_tgw_attachment == "public")? 1 : 0}"
+  transit_gateway_id = "${aws_ec2_transit_gateway.this.id}"
+  tags               = "${merge(map("Name", format("%s", var.name)), var.tags, var.tgw_tags)}"
+
+  depends_on = ["aws_ec2_transit_gateway.this"]
+}
+
+#########
+# Transit Gateway Route table association and propagation for TGW Attachements
+#########
+
+resource "aws_ec2_transit_gateway_route_table_association" "this" {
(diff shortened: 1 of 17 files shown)


main.tf
```diff
@@ -906,3 +906,77 @@ resource "aws_default_vpc" "this" {
...
+  transit_gateway_route_table_id = "${aws_ec2_transit_gateway_route_table.this.id}"
+  depends_on                     = ["aws_ec2_transit_gateway.this"]
+}
+
+########
+# Routes to TGW are added to the Subnets.
+########
+
+locals {
+  route_table_id = ["${split(",", var.subnet_type_tgw_attachment == "private" ? join(",", aws_route_table.private.*.id ) : join(",", aws_route_table.public.id))}"]
```
robh007: The public route table produces an error on the join as it's a string and not a list. 
```terraformhcl
aws_route_table.public.id needs to be aws_route_table.public.*.id
```

ebalakumar: Hello @robh007  . Thanks for finding it out . I have resolved the same . Please verify.

robh007:
I've been using TGW and working through the workflow. This commit IMHO doesn't really help with configuring and setting up to use transit. There's a number of outside influences that are required for transit to work. I almost think it's a separate module. With this current way of thinking all you can do is create a transit gateway and attach to a single VPC. It sort of defeats the idea of using a transit gateway. You can use a single transit with multiple routing domains. If this module was to include transit it needs to cater for existing TGWs so ```${var.create_tgw}``` possibly needs to only influence the **creation of the gateway**. There should be some other conditions to attach a gateway, create or attach transit gateway route tables & association to route tables.

Example 

- VPC 1 - Creates TGW
- VPC 2 - Attaches to TGW created by VPC 1 and maybe creates a separate routing domain
- VPC 3 - Attaches to an existing TGW and an already existing routing domain.

For Multiple Accounts (Organisations) I'd be looking to use resource sharing to share the TGW to other accounts. Then attach and create where necessary, also VPN attachment to TGW & DX attachment when supported. 

Happy to help. I'm actively building something now so hence my comments as I was trying to work with your branch of this module.

ebalakumar:
Hello , 

The Reason for this approach :
1. The current VPC module creates one VPC in a single account.
2. Based on discussion https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/207 . Its discussed to have a single module which can do TGW as well . 

I would love to have TGW created in a separate module and I did thought of all aspects which you mentioned . But I tried to stick this module which its currently capable . 

Looking forward to see your commits . Will work together . Thanks.

antonbabenko:
Just my 5 cents... I think we need to have TGW resources to be as part of this module AND as a separate module, but starting by extending this module is a good idea.

StephenKing:
Think it makes sense to offer attachment to an existing TGW (plus adding routes to all route tables).

robh007:
@ebalakumar I've created a PR from your original and pushed to your repo. It adds the ability to attach an existing TGW & also using the AWS RAM service share a TGW to another account. I've also updated the examples to include a simple VPC hub & spoke. This creates the TGW and two VPCs in the same account. Also a cross account example for a VPC Hub and Spoke using RAM. 

There also appears to be a bug on destroy, but running destroy a second time works fine. I've not investigated properly, but the route table IDs appear to be duplicated. 

@antonbabenko I've also updated the variables to have types for list(s) & map(s). Don't know if they've been left out intentionally. But the HCL plugin was complaining so I removed all the red :-).

ebalakumar:
Hello @antonbabenko  . Please take a look and share your comments .

ebalakumar:
> @ebalakumar I've created a PR from your original and pushed to your repo. It adds the ability to attach an existing TGW & also using the AWS RAM service share a TGW to another account. I've also updated the examples to include a simple VPC hub & spoke. This creates the TGW and two VPCs in the same account. Also a cross account example for a VPC Hub and Spoke using RAM.
> 
> There also appears to be a bug on destroy, but running destroy a second time works fine. I've not investigated properly, but the route table IDs appear to be duplicated.
> 
> @antonbabenko I've also updated the variables to have types for list(s) & map(s). Don't know if they've been left out intentionally. But the HCL plugin was complaining so I removed all the red :-).

Merged .

ebalakumar:
Hello @robh007 , When using existing TGW , the step for association and propagation of TGW Attachment in the TGW route table is missed . It was creating only the TGW Attachment . I have resolved it .

robh007:
Hi @ebalakumar that's odd. My local copy does have that code in, anyway you've resolved it.

ebalakumar:
Hello @robh007 , With the current setup . We can achieve the below with respect to TGW.
1. Create VPC and attach it to a new TGW .
2. Create VPC and attach it to an existing TGW . 

I have another scenerio .
3. Create VPC-2 , attach it to an existing TGW which is already attached with VPC-1 , Updates the TGW route table and subnet routes so that  VPC 1 and VPC 2 can communicate .
 -- > The current setup is not able to do that , as while attaching existing TGW , it actually creates new route table for the TGW . In this case , we have to use the existing route table . I am creating a new branch for this . We can discuss about this further and decide if its useful.

ebalakumar:
> There also appears to be a bug on destroy, but running destroy a second time works fine. I've not investigated properly, but the route table IDs appear to be duplicated.
> 

The issue is resolved . It was happening due to count calculation in tgw_route resource.

ebalakumar:
Hello @antonbabenko , Do you have any comments and suggestions .

JoshiiSinfield:
Hi @ebalakumar @robh007 ,

This PR fails when you set ` create_vpc = false ` to create just a TGW separately from the vpc creation.

image

due to there being no vpc resource for the below:
image

JoshiiSinfield:
Sorry for the abrupt last comment, I was in a rush yesterday and didn't want to forget about this!.

I believe the error is due to `vpc_id = "${aws_vpc.this.id}"`. 
If create_vpc = false there is no vpc resource for it to get the id from. At a minimum I think this needs to be changed to:
`vpc_id = "${element(concat(aws_vpc.this.*.id, list("")), 0)}"`
That should remove the issue. I will update the code and test it shortly.

Also, after trying this PR out further I personally feel that the TGW resources should be split out into a sub module, which this main.tf can make use of itself.  Doing this will allow simple setups inline with VPCs to continue working here, but also allow more complex setups outside of VPC creation etc. to be managed without any dependency on the rest of the vpc workings.

I'd be happy to take a shot at this and push a commit up if people think it's a good idea?

Cheers,
Josh

StephenKing:
If you want to hear my opinion: TGW creation is nothing that belongs into this module. Attaching it to an existing one and setting up the routing tables fits well.

robh007:
@StephenKing I do agree with you. I did mention this above in my initial comments before I did some of the initial work which formed part of this PR.

DrFaust92:
I agree with the approach of moving the TGW creation to a separate module and created a repo in the spirit of the VPC module

https://github.com/DrFaust92/terraform-aws-tgw 

take a look and let me know what you think.

robh007:
I did create a separate module for the TGW work and published it on the terraform registry. https://registry.terraform.io/modules/BorisLabs/transit-gateway/aws/0.0.1

antonbabenko:
I have pushed TGW resources into a separate module - https://github.com/terraform-aws-modules/terraform-aws-transit-gateway, and verified the integration with https://github.com/terraform-aws-modules/terraform-aws-vpn-gateway.

I am not sure if there is anything else missing in this VPC module in regards to integration with TGW. I like the idea that TGW is created separately and not as part of the VPC module.

As @StephenKing said in his comment: "Attaching it to an existing one and setting up the routing tables fits well."... Do you guys think that this is a good idea to allow the management of `aws_ec2_transit_gateway_vpc_attachment` and routes in both modules? I'd prefer to leave it as is for now, since it is rather tricky to do right in a compatible way.

antonbabenko:
Closing this PR. TGW lives in a separate module.

PR #242, opened 2019-04-02
Add Output Of Subnet ARNs

blaines:
Facilitates resource access manager, subnet sharing across accounts

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,7 @@ Terraform version 0.10.3 or newer is required for this module to work.
 | database\_route\_table\_ids | List of IDs of database route tables |
 | database\_subnet\_group | ID of database subnet group |
 | database\_subnets | List of IDs of database subnets |
+| database\_subnet\_arns | List of ARNs of database subnets |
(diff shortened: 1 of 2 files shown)


antonbabenko:
@blaines Looks good to me. Do you think it makes sense to output ARNs of all resources? (I have not worked with RAM myself, so not sure how it works in details).

blaines:
Maybe, right now it looks like RAM is for Subnets, Transit Gateways, and Resolver Rules. So I think there's nothing else to add at this time. (This doesn't to Transit right?)

antonbabenko:
@blaines Thanks for this PR. I think you are right, you have added ARNs to all resources managed by this module.

[v1.64.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.64.0) has been released.

dyna-dot:
[CHANGELOG.zip](https://github.com/terraform-aws-modules/terraform-aws-vpc/files/3136379/CHANGELOG.zip)

PR #243, opened 2019-04-08
add support for KMS VPC endpoint creation

nloutas:
also add missing list and map types in certain variables


antonbabenko:
Thank you. Merged, but there was no need to add missing types (at least not in the same PR).

[v1.62.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.62.0) has been released.

PR #245, opened 2019-04-20
Added ARN of VPC in module output

rafilkmp3:
I need in my policy generator the arn of vpc so I would like to include this


antonbabenko:
Thanks!

[v1.63.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.63.0) has been released.

PR #248, opened 2019-04-29
Add SQS Endpoint support

jvelasquezjs:


jeohist:
This is also a good template for other services which require a VPC Interface Endpoint (such as Kinesis Firehose) :-)

antonbabenko:
Thank you for adding this feature.

I've just released [v1.66.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.66.0) with this feature.

PR #252, opened 2019-05-06
Allow ICMP Network ACL rules

donovancade:
Updating VPC module to allow ICMP Network ACL rules.  Since from_port and to_port are ignored for ICMP rules, allowing user to leave these fields off makes sense.  We could make this more complicated and require from_port and to_port if protocol is not ICMP but that would make the module less readable.


antonbabenko:
Hi @donovancade !

Thanks for adding this feature.

[v1.70.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v1.70.0) has been released.

PR #270, opened 2019-05-26
Upgrade Docker Image to fix CI

Skaronator:
Not sure if this is enough to fix CI.


antonbabenko:
I am not paying almost any attention to CI checks for a rather long time, so thanks for an update!

PR #275, opened 2019-06-01
Fix opportunity to create the vpc

FKT-psw-circleci:
Hi there,
I found out, that avoid creation of VPC leads to creation of public subnets.
This fix will avoid the creation if "create_vpc = false".
Best regards
Sven

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 terraform.tfstate
 terraform.tfvars
 Gemfile.lock
+.idea
--- a/main.tf
+++ b/main.tf
@@ -248,7 +248,7 @@ resource "aws_route_table" "intra" {
 # Public subnet
 ################
 resource "aws_subnet" "public" {
-  count = var.create_vpc && length(var.public_subnets) > 0 && false == var.one_nat_gateway_per_az || length(var.public_subnets) >= length(var.azs) ? length(var.public_subnets) : 0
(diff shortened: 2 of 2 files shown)


SvenRedelin1601:
Hi @antonbabenko ,
To provide some small changes to vpc module project is a pull-request the right way?
I would be very happy, if you can review and merge this small change, which might be really necessary if you want to avoid vpc creation.
Thanks and regards
Sven

antonbabenko:
Thanks for this PR! I have opened related issue in Terraform core - https://github.com/hashicorp/terraform/issues/21592 - and I will review if there is a similar problem in other parts of this module during an hour.

Big thanks for this discovery!

@SvenRedelin1601 Submitting PRs is always a great idea.

antonbabenko:
Fixed in b253c94c6774270b100d419102de5d5d9c425c00 . There are couple more resources affected by this bug.

[v2.3.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.3.0) has been released.

.gitignore
```diff
@@ -4,3 +4,4 @@
 terraform.tfstate
 terraform.tfvars
 Gemfile.lock
+.idea
```
antonbabenko: Consider using [global .gitignore](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore) for dirs like `.idea`.

PR #276, opened 2019-06-03
Adds support to create a subnet to be used by a VPC based elasticsearch domain

jeffsteinmetz:
Adds support to create a subnet to be used by a VPC based elasticsearch domain.
(Does not create the elasticsearch domain, just the subnets)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ These types of resources are supported:
 * [DHCP Options Set](https://www.terraform.io/docs/providers/aws/r/vpc_dhcp_options.html)
 * [Default VPC](https://www.terraform.io/docs/providers/aws/r/default_vpc.html)
 * [Default Network ACL](https://www.terraform.io/docs/providers/aws/r/default_network_acl.html)
+* Elasticsearch Subnet - A set of subnets within a VPC where a VPC based elasticserach domain can be launched.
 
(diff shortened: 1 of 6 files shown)


jeffsteinmetz:
Example of the result after a `terraform apply` using this PR

Subnets___VPC_Management_Console

antonbabenko:
Jeff, I am going to close this PR because there is no such type of subnet as "elasticsearch" in AWS. It can be any type of subnet (public or private), and this module already allows to create such generic types of subnets.

jeffsteinmetz:
Thank you, I am now wondering with this existing module, can you create multiple private subnets with different naming schemes?
The idea would be to create a private subnet for EC2 docker images, and a "different" private subnet for use with elasticsearch.
The goal was to create another subnet, isolated from any other private subnets.

antonbabenko:
Currently, this module does not have support for different naming schemas but it is possible to use a different length of `azs` and `private_subnets`, for example, look at this example: https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/examples/issue-44-asymmetric-private-subnets/main.tf#L15-L18

Also, you can create subnets as `private_subnets` using this module and use a slice of them for different purposes in your code. Eg, values with index 0-2 - private; index 2-4 - elasticsearch.

Issue #250, opened 2019-05-03
Could we add the option to use a NAT instance rather than a NAT gateway in the VPC?


bmacauley-reward:
NAT gateways start to get expensive when you have many accounts and VPC's

Would it be possible to add the option to use the older NAT instance instead of the NAT gateway? If had the option to set the instance size as well, we could scale down to micro's or nano's

https://www.theguild.nl/cost-saving-with-nat-instances/

Jetbrains example...
https://github.com/jetbrains-infra/terraform-aws-vpc-with-private-subnets-and-nat

sirket:
You may be better off simply using single_nat_gateway = true rather than running NAT instances. An undersized NAT instance coupled with a developer who suddenly decides to push a bunch of traffic can result in dropped traffic and hard to troubleshoot problems. A single NAT gateway won't have that issue while also saving money. For a dev or testing account the loss of redundancy shouldn't pose any problems. For production accounts the problems with NAT instances outweigh any cost savings.

bmacauley-reward:
Agreed... NAT gateways for production

But i would like NAT instances for non-prod....with the ability to turn them off at scheduled times to save cost

bonovoxly:
Custom configurations may require a NAT instance. This would be a useful feature.

antonbabenko:
This is out of scope for this module as mentioned in #281 

Closing this issue.

PR #281, opened 2019-06-06
Added basic support for an Ubuntu 18.04 NAT instance gateway

bonovoxly:
# Description

This PR attempts to address #250 .

It adds support for a NAT instance gateway, using Ubuntu 18.04 as the AMI. Note that it requires an AWS keypair to be created and named first. 

Hopefully the conditionals are sound, I did test back and forth going from an AWS managed NAT gateway to an instance, and it seemed to work well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,24 @@ This module supports three scenarios for creating NAT gateways. Each will be exp
 
 If both `single_nat_gateway` and `one_nat_gateway_per_az` are set to `true`, then `single_nat_gateway` takes precedence.
 
+## NAT Instance Gateway
+
+This module also supports the use of an Ubuntu 18.04 NAT instance gateway. To use a NAT instance:
+
+* Create a keypair, and upload that to EC2. Note the name.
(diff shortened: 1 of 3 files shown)


bonovoxly:
I'm going to close this; it opens a pandoras box of things to deal with that might not be worth it.

antonbabenko:
Bill, if you have time and interest, I think it would be a better idea to improve [this NAT instance module](https://github.com/terraform-community-modules/tf_aws_nat) which we already have, add support for integration with it and document in details with examples how this can be done.

bonovoxly:
Roger that, @antonbabenko . It's a bummer, I found a fairly clean solution, but I worry that the amount of options that someone may want.

Regardless, I can reopen the PR if there's enough demand for a simple ubuntu nat gateway.

antonbabenko:
I agree, there are lots of details and options always.

No need to open if you are not going to work on it. _I am chasing and closing/resolving issues. Fewer issues = less work for me._ :)

PR #297, opened 2019-06-27
Abstract simple-vpc module attributes into variables with defaults

sover02:
# Description

The simple-vpc example is super useful! I know it's meant to be an example, but users may not want the default availability zones, or a couple of other variables so I abstracted them into variables.

There is no change to deployment if variables aren't specified.

### DIFF
--- a/examples/simple-vpc/main.tf
+++ b/examples/simple-vpc/main.tf
@@ -10,11 +10,11 @@ data "aws_security_group" "default" {
 module "vpc" {
   source = "../../"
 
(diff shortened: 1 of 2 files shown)


antonbabenko:
Thanks for the PR, but the main reason why there are very small usage of variables, locals and other features in examples is to make it obvious for users to know what values are expected and let it up to a user to decide how to provide them.

sover02:
Faire enough, makes sense.

PR #299, opened 2019-07-05
feat: tags templating

antonu17:
# Description

Firstly, thanks a lot for a great job on the module!
This is more like a feature request rather than a pull-request at this point.
Background: there were a bunch of VPCs created by different tools, including kops. At some point, we decided to import all VPCs resources to a single terraform configuration, using this marvelous module. But we have a craving to keep some of our tags including "Name" matching certain format.

Here is a naive implementation in 0.12 syntax which allows us to define specific resources' tags like this

    private_route_table_tags = {
      "Name"                    = "private-%az%.%name%"
      "kubernetes.io/kops/role" = "private-%az%"
    }

Does it make sense? Has anyone else faced the same needs?


gillg:
Seems to need to be closed...

PR #300, opened 2019-07-08
Add IPv6 support (Terraform 0.12 edition)

StephenKing:
Add variable `enable_ipv6` to allow enabling IPv6 support (resulting in passing `assign_generated_ipv6_cidr_block` to `aws_vpc`).

Enabling IPv6 support further results in an egress-only internet gateway being provisioned and routing tables of subnets being adjusted.

Additional variables allow to choose the indices out of the /64 subnet based on the assigned /56 range.

This is based on @bcenker's work in #16 as well as @dekimsey's on #218, but based on a clean master branch (after the Terraform 0.12 upgrade).

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -21,36 +21,36 @@ variable "assign_generated_ipv6_cidr_block" {
 }
 
 variable "enable_ipv6" {
-  description = "Assigns ipv6 subnets and routes"
+  description = "Assigns IPv6 subnets and routes"
   default     = false
 }
 
 variable "private_subnet_ipv6_prefixes" {
-  description = "Assigns ipv6 subnet id based on the Amazon provided /56 prefix base 10 integer (0-256). Must be of equal length to the corresponding ipv4 subnet list"
+  description = "Assigns IPv6 subnet id based on the Amazon provided /56 prefix base 10 integer (0-256). Must be of equal length to the corresponding IPv4 subnet list"
   default     = []
   type        = "list"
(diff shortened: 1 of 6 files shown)


variables.tf
```diff
@@ -20,6 +20,40 @@ variable "assign_generated_ipv6_cidr_block" {
   default     = false
 }
 
+variable "enable_ipv6" {
+  description = "Assigns IPv6 subnets and routes"
+  default     = false
+}
+
+variable "private_subnet_ipv6_prefixes" {
```
StephenKing: Strictly speaking.. those are not prefixes, but more the.. indices of the prefixes? Suggestions?

dekimsey: Agreed, they aren't. They are the last octet of the given ipv6 prefix. IPv6 docs tends to be prefix based, so I chose that. Admittedly, it's not a great name, but I couldn't for the life of me figure out what a better way to describe the fact they are an enumerated list of the last octet in an IPv6 prefix such as `2001:db8::ff00:42:83XX`. I explicitly didn't go with index or slice because I didn't want it to be confused as an index of another list.

So maybe `..._subnet_ipv6_octet`?

StephenKing: What about giving an example in the description? While it already suggests 0-256, an example like `[0, 1, 2]`might make it clearer. The description (and the topic) feels pretty complex :-)

antonbabenko:

PR #303, opened 2019-07-20
Add vpc name

hootieben:
# Addition of Name Output And CircleCI Fixes

- Added vpc_name as an output
- Modified circleci checks:
  - Install missing curl binary for tflint checks
  - Add default aws region required by aws module
  - Modified init and validate steps to traverse directories
  - Removed invalid (in 0.12.x) -check-variables flag

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -624,7 +624,7 @@ output "azs" {
   value       = var.azs
 }
 
-output "vpc_name" {
+output "name" {
   description = "The name of the VPC"
   value       = var.name
 }
--- a/outputs.tf
+++ b/outputs.tf
@@ -624,3 +624,7 @@ output "azs" {
   value       = var.azs
 }
 
+output "vpc_name" {
+  description = "The name of the VPC"
+  value       = var.name
+}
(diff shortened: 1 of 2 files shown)


outputs.tf
```diff
@@ -624,3 +624,7 @@ output "azs" {
   value       = var.azs
 }
 
+output "vpc_name" {
+  description = "The name of the VPC"
+  value       = var.name
```
antonbabenko: I don't think there is `name` as an attribute here - https://www.terraform.io/docs/providers/aws/r/vpc.html

hootieben: Correct, it's not an exported reference from the aws provider, it's something that's being pulled from the input variables (like AZS above it there). Our organization has a small number of VPCs so we've found it helpful to reference the name from a remote_state lookup.

antonbabenko: Right, let's call output `name` instead of `vpc_name`. Make a change and I will merge it right away.

hootieben: Ahh yes, you're correct, that's the better method. I'll make that change.

antonbabenko:
Thanks, Ben!

[v2.9.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.9.0) has been just released.

Issue #259, opened 2019-05-17
Feature Request: Specific Subnet tag of same type subnets


morihaya:
Hi there. Thank you for great module "terraform-aws-vpc" , It's helped me. so thanks.
I think, it's great if you can tag subnets more finely like below.

```
private_subnet_tags = {
  type = "map"

  subnet[0] = {
    "kubernetes.io/cluster/eks-dev"     = "share"
  }
  subnet[1] = {
    "kubernetes.io/cluster/eks-stg"     = "share"
  }
  subnet[3] = {
    "kubernetes.io/cluster/eks-prd"     = "share"
  }
}
```

mwthink:
+1 This is crucial for running multiple EKS clusters in the same VPC provisioned by this module, less diffs always be found or implementing some other "creative" solutions.

mwthink:
Alternatively, the module could accept `eks_subnets` and a matching list of `eks_cluster_names` to achieve the same result. 

For anyone unfamiliar with EKS, there are some simple tags that must be applied to subnets for it to work properly (see them [here](https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html)). EKS itself will add these tags itself when provisioning (At least to the private subnets it's provided), but this causes terraform to detect diffs (And delete the tags) every single time.

antonbabenko:
@morihaya Thanks for this idea and info regarding EKS expectations (@mwthink). I think this should be rather possible to achieve using Terraform 0.12.

If anyone can write at least the initial pull-request it would be great.

I can't guarantee when I will have time to prioritize this feature myself because I am busy with all other upgrades of terraform-aws-modules to 0.12.

skang0601:
I'll take a crack at this later this week. This is something I"m going to need for my use case.

skang0601:
So I was thinking a little about this, would something simple as this suffice?
```resource "aws_subnet" "private" {
  count = var.create_vpc && length(var.private_subnets) > 0 ? length(var.private_subnets) : 0

  vpc_id            = local.vpc_id
  cidr_block        = var.private_subnets[count.index]
  availability_zone = element(var.azs, count.index)

  tags = merge(
    {
      "Name" = format(
        "%s-${var.private_subnet_suffix}-%s",
        var.name,
        element(var.azs, count.index),
      )
    },
    var.tags,
    lookup(var.private_subnet_tags, "global", {}),
    lookup(var.private_subnet_tags, count.index, {}),
  )
[...]

PR #304, opened 2019-07-23
allow private/public subnets to be individually tagged

skang0601:
# Description

Changed the private & public subnet tags to be a complex map. This allows private and public subnets to be individually tagged.

Addresses #259

### DIFF
--- a/examples/eks-subnet-tagging/main.tf
+++ b/examples/eks-subnet-tagging/main.tf
@@ -1,0 +1,31 @@
+module "vpc" {
+  source = "../../"
+
+  name = "my-eks-cluster"
+
+  cidr = "10.0.0.0/16"
+
+  azs             = ["eu-west-1a", "eu-west-1b", "eu-west-1c"]
+  private_subnets = ["10.0.0.0/19", "10.0.32.0/19", "10.0.64.0/19", "10.0.128.0/20", "10.0.144.0/20"]
+  public_subnets  = ["10.0.160.0/19", "10.0.192.0/19", "10.0.224.0/20", "10.0.240.0/20"]
+
+  enable_nat_gateway = true
+  single_nat_gateway = true
+
+  private_subnet_tags = {
+    global = { foo = "bar" }
+    0 = { "kubernetes.io/cluster/my-eks-cluster-name" = "shared" }
+    1 = { "kubernetes.io/cluster/my-eks-cluster-name" = "shared" }
+    2 = { "kubernetes.io/cluster/my-eks-cluster-name" = "shared" }
+    3 = { "kubernetes.io/role/internal-elb" = 1 }
+    4 = { "component" = "some other component"}
+  }
+
+  public_subnet_tags = {
+    global = { foo = "bar" }
+    0 = { "kubernetes.io/cluster/my-eks-cluster-name" = "shared" }
+    1 = { "kubernetes.io/cluster/my-eks-cluster-name" = "shared" }
+    2 = { "kubernetes.io/cluster/my-eks-cluster-name" = "shared" }
+    3 = { "component" = "some other component"}
+  }
+}
--- a/examples/network-acls/main.tf
+++ b/examples/network-acls/main.tf
@@ -32,7 +32,7 @@ module "vpc" {
   single_nat_gateway = true
 
   public_subnet_tags = {
-    Name = "overridden-name-public"
+    global = { Name = "overridden-name-public" }
   }
 
   tags = {
--- a/examples/secondary-cidr-blocks/main.tf
+++ b/examples/secondary-cidr-blocks/main.tf
@@ -19,7 +19,7 @@ module "vpc" {
   single_nat_gateway               = true
 
   public_subnet_tags = {
-    Name = "overridden-name-public"
+    global = { Name = "overridden-name-public"}
   }
 
   tags = {
--- a/main.tf
+++ b/main.tf
@@ -16,6 +16,18 @@ locals {
     ),
     0,
   )
+
+  private_subnet_tags = merge(
+    { "global" = {} },
+    { for i in range(length(var.private_subnets)): i => {} },
+    var.private_subnet_tags
+  )
+
+  public_subnet_tags = merge(
+    { "global" = {} },
+    { for i in range(length(var.public_subnets)): i => {} },
+    var.public_subnet_tags
+  )
 }
 
 ######
@@ -264,7 +276,8 @@ resource "aws_subnet" "public" {
       )
     },
     var.tags,
-    var.public_subnet_tags,
+    local.public_subnet_tags["global"],
+    local.public_subnet_tags[count.index],
   )
 }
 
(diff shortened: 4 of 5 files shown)


bmcstdio:
@mwthink @antonbabenko is there a way I can help in order to get this in soon?

skang0601:
@mwthink @antonbabenko Any chance this can get reviewed? Working off a forked version to run a cluster in it atm.

antonbabenko:
@skang0601 Thanks for the PR!

This PR requires some time to have a deeper look from my side (mostly because this is backward incompatible change) and I am on vacation now. I will be back to this PR by the end of August.

ozbillwang:
@morihaya & @skang0601 

Your case is very special and not best practices.

For example, how should you allow production to run in one subnet only?

I dont' recommend any PRs for this type of use case.

lmserrano:
@ozbillwang , I think that regardless how the requester intends to make use of these features, it still makes sense to me to be able to assign tags with fine granularity to each subnet, because the official Terraform AWS Provider for `aws_subnet` supports this, and also because it is legitimate to want to have for example 3 subnets and have tag A for first, B for second, C for third, for whichever reason or purpose, from human to any kind of script or automation requirements

ozbillwang:
Can you give any useful case that you have to make different tags to private subnets or public subents? 

Second, resource of `aws_subnet` will of course allow tags for each resource. But how can you ask a module to use `count` and add the useless complexity to allow the subnets to be different tags.

For example, will you do different TAGs to ec2 instances, when you set autoscaling to scale the ec2? You of course can say the resource `aws_ec2_instance` supports fine grain tag for each ec2.

ozbillwang:
@lmserrano

skang0601:
@ozbillwang 
The comparison to ec2 instances in an asg is less apt since by design instances in an asg is homogenous. Subnets in a VPC is less so since several different environments/tenants/components can reside in a VPC.

> For example, how should you allow production to run in one subnet only?
By using the aws subnet datasource to query for the specific subnet.

I do something similar for components rather than environments by doing something like this:

```
private_subnet_tags = {
    global = { "Tier" = "private"}
    0 = { "service" = "rabbitmq", "component" = "app"}
    1 = { "service" = "rabbitmq", "component" = "app" }
    2 = { "service" = "rabbitmq", "component" = "app" }
    3 = { "service" = "elasticsearch", "component" = "app" }
    4 = { "service" = "elasticsearch", "component" = "app" }
    5 = { "service" = "elasticsearch", "component" = "app" }
}
...

data "aws_subnet_ids" "private" {
  tags = {
    service = "rabbitmq"
    Environment = var.environment
    Tier = "private"
    component = "app"
  }
  vpc_id = data.aws_vpc.vpc.id
}
```

This also can give me the flexibility of mapping out different parts of the vpc to different tenants/teams.

lmserrano:
@ozbillwang , I think @skang0601 has provided a good example of this (thank you for the example @skang0601 ). In short, when you have multiple subnets (and it's just not about private and public), it can often be useful to have a way to easily identify properties in them in a human-readable or machine-unequivocal way. Also, the set of subnets you have in each of your VPCs is typically not something that changes automatically with scaling, but something you set, based on your needs (and which requires you to consider in advance the ip ranges you have available and the number of instances you expect to have in each subnet), thus which you should be able to label it in advance, as I see it.

ozbillwang:
@skang0601 

Your example is another sample of useless complexity, which you can set as below ways
```
private_subnet_tags = {
    "service" = "rabbitmq,elasticsearch"
    component" = "app"
}
```

skang0601:
@ozbillwang I don't completely follow what your example accomplishes? The point of individually tagging subnets is so I can place different components in different subnets so I can define things like egress/ingress control in a fine grain way. 
It's oftentimes I can't use something like security groups because:
  1) They exist in different VPCs which are peered across regions
  2) I don't own the security group but only have knowledge about their ip space.

I do not want a situation where granting access to my rabbit cluster may inadvertently grant access to my elasticsearch cluster and so on. And it also can come in handy when I'm debugging since I should know the range of IPs resources are in.

RyPeck:
Perhaps we could leave `private_subnet_tags` as is and add an additional input such as `private_subnet_tags_specific` that allows you to specify specific subnets in the list as you have done. That way we preserve backward compatibility while adding the new feature.

I would also suggest being able to specify subnets 0 - 2 and 3 - 5.

```
private_subnet_tags = {
  foo = "bar"
}

private_subnet_tags_specific  = {
  [0, 2] = { "service" = "rabbitmq", "component" = "app"}
  [3, 5] = { "service" = "elasticsearch", "component" = "app" }
}
```

bmcstdio:
I would very much like to see this going in some time soon, and I'd be happy to help testing.

> Can you give any useful case that you have to make different tags to private subnets or public subents?

@ozbillwang another use case is being able to use [subnet auto discovery](https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/controller/config/#subnet-auto-discovery) with the [AWS ALB ingress controller](https://github.com/kubernetes-sigs/aws-alb-ingress-controller) in an EKS cluster. Currently, and since subnets cannot be individually tagged, there's no way to set `kubernetes.io/role/internal-elb` and/or `kubernetes.io/role/elb` according to the desired topology.

skang0601:
@RyPeck Had a chance to look at this again, I don't think there's a real good way of accomplishing what you want without doing something incredibly messy. 

Perhaps I can meet somewhere in the middle and retain the current behavior of private_subnet_tags and have another variable for the specific tags as I have them now.

ozbillwang:
@bmcstdio 

Finally I understand where the problem you have. 

For the last sample you gave, in fact, you mixed public subnets and private subnets. The exist codes do have the feature to have different tags on public subnets or private subnets already. 

Below codes I used for a while with this module. 

```
  private_subnet_tags = {
    "kubernetes.io/cluster/eks-sit-cluster" = "shared"
    "kubernetes.io/role/internal-elb" = "1"
  }

  public_subnet_tags = {
    "kubernetes.io/role/elb" = "1"
  }
```

gillg:
I'm very not aggreed this idea. Moreover, you can easly workaround in your context using the native subnet_tags vars...

DrFaust92:
as an interim solution, use https://www.terraform.io/docs/providers/aws/r/ec2_tag.html

isac:
I would really like to see this feature.
My use case is also for EKS.

I want to add EKS specific tags to all except two of the subnets in my VPC.

rodmoreno:
Yeah I'm trying with same scenario that @isac because my VPC is on us-east-1 with 6 subnets but I want to only tag first 4 (EKS doesn't support my last 2 subnets availability zones)...

paultung:
I am extremely looking forward this feature.
Hopefully could be merge to `master` as soon as possible.

e-tip:
Another use case is the one I'm working on right now.
I have two EKS clusters on the same VPC, each one of the clusters has a load-balancer-controller. I've created 4 subnets, and I want the load-balancer-controller on A cluster to work on the first two subnets, and the load balancer controller on cluster B works on the second two subnets. Without single tagging this is not possible with subnet autodiscovery

PR #305, opened 2019-07-25
[WIP] Ignore tag changes for vpc and subnets

hp197:
More and more services in aws (for example EKS) create tags on existing resources for proper working. Upon Terraform run, those tags are removed as they where not in the original template.
This patch is a tradeoff between usability and user friendliness.

After applying this patch it will create the tags who are defined in the landscape.tf but who are not yet in the state file. If the tag is not defined in the landscape file it wont remove it nor will it try to change tags if it is changed in aws but not in the landscape file. Once set they are now unmanaged.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -66,6 +66,10 @@ resource "aws_vpc_dhcp_options" "this" {
     var.tags,
     var.dhcp_options_tags,
   )
+
+  lifecycle {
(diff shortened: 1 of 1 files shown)


hp197:
Please wait for the WIP tag before merging, might be extra commit inbound.

PR #306, opened 2019-07-26
Fix removal of auto generated aws tags

hp197:
More and more services in aws (for example EKS) automatically create tags on existing resources for proper working. Upon Terraform run, those tags are removed as they where not in the original template.

After applying this patch it will create the tags who are defined in the landscape.tf but who are not yet in the state file. If the tag is not defined in the landscape file it wont remove it nor will it try to change tags if it is changed in aws but not in the landscape file. Once set they are now unmanaged.

It allows for AWS to autocreate the tag:
```kubernetes.io/cluster/<cluster-name> = "shared"```
on the vpc and subnets used in EKS.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -37,6 +37,10 @@ resource "aws_vpc" "this" {
     var.tags,
     var.vpc_tags,
   )
+
+  lifecycle {
+    ignore_changes = [tags]
+  }
 }
 
 resource "aws_vpc_ipv4_cidr_block_association" "this" {
@@ -66,6 +70,10 @@ resource "aws_vpc_dhcp_options" "this" {
     var.tags,
     var.dhcp_options_tags,
   )
+
+  lifecycle {
+    ignore_changes = [tags]
+  }
 }
 
 ###############################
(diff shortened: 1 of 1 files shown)


antonbabenko:
Hi Henry,

This would be just great and easy if it worked for all cases.

Terraform does not allow to specify which tags should be ignored, but we want to have a way to tag resources and manage tags as code.

Closing this until Terraform has support for variables in `ignore_changes` lifecycle.

hp197:
@antonbabenko 

So what you are basically are saying is that we shouldn't use the EKS service with this terraform module, as with EKS the VPC is tagged with tags who are mandatory for the workings of the EKS cluster.
Who breaks the working of who?

antonbabenko:
Please check this example of how both modules can be integrated: https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/basic/main.tf#L85-L111

hp197:
Yes, and this work if you make your VPC in the same run as your EKS cluster. But for us the VPC is created in a separate run and the EKS cluster(s) are created in a separate run (so VPC, EKS, EC2, etc, etc are all decoupled with us). Currently there is no way in terraform to add tags to a yet existing VPC.

sc250024:
@hp197 I have the same setup as you; separate VPC and EKS modules within a Terraform repo. For EKS, since I have a standard naming convention, I know what the clusters should be named, so I put those EKS VPC tags in my VPC module after the fact.

PR #309, opened 2019-07-29
remove extra unused route

tushar00jain:
The database route "aws_route.database_nat_gateway" is not utilised in any route table association and is a duplicate of aws_route.private_nat_gateway

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -181,18 +181,6 @@ resource "aws_route" "database_internet_gateway" {
   }
 }
 
-resource "aws_route" "database_nat_gateway" {
-  count = var.create_vpc && var.create_database_subnet_route_table && length(var.database_subnets) > 0 && false == var.create_database_internet_gateway_route && var.create_database_nat_gateway_route && var.enable_nat_gateway ? local.nat_gateway_count : 0
-
-  route_table_id         = element(aws_route_table.private.*.id, count.index)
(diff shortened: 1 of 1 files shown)


antonbabenko:
This is used when a user wants to route database subnets via NAT gateway.

I think so :)

bryantbiggs:
Though it may not be used by many, it still is a valid resource to provide NAT translation for traffic originating in the database subnets and therefore it will be kept in scope of the module

bmickunas:
I think this is actually bug rather than useless. Shouldn't the `route_table_id` parameter reference the database route table instead of the private route tables?

So change:
`route_table_id         = element(aws_route_table.private.*.id, count.index)`

to this:
`route_table_id         = element(aws_route_table.database.*.id, count.index)`

DrFaust92:
@bmickunas, i think you are right. @antonbabenko/ @bryantbiggs thoughts?

i would consider a dedicated nat for database subnet, otherwise it makes the private subnet more of a proxy in this case (which is fine)

bmickunas:
I have a project using 2.0.0. How should I go about creating a tag for v2.0.1 with the same fix I have in #476?

antonbabenko:
This has been fixed in #476, and [v2.48.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.48.0) has been just released.

Thanks everyone!

Issue #23, opened 2017-11-11
Feature request - VPC flow log


antonbabenko:
Inspired by https://github.com/terraform-community-modules/tf_aws_vpc/issues/64 and https://github.com/GSA/terraform-vpc-flow-log

Some points to take into account when adding VPC flow log feature to this VPC module:
- [ ] Make creation of [aws_flow_log](https://github.com/GSA/terraform-vpc-flow-log/blob/master/flow_log.tf) conditional. Disabled by default.
- [ ] Make creation of [aws_cloudwatch_log_group](https://github.com/GSA/terraform-vpc-flow-log/blob/master/iam.tf) conditional. User should be able to provide existing log group.
- [ ] Limit resources in [log_policy](https://github.com/GSA/terraform-vpc-flow-log/blob/master/log_policy.json) to only what is necessary.
- [ ] Make creation of [aws_iam_role](https://github.com/GSA/terraform-vpc-flow-log/blob/master/iam.tf) conditional. User should be able to provide existing IAM role.
- [ ] Do not create `aws_cloudwatch_log_group` and `aws_iam_role` resources, if `aws_flow_log` is disabled.
- [ ] Make changes to an existing [complete-vpc example](https://github.com/terraform-aws-modules/terraform-aws-vpc/tree/master/examples/complete-vpc)
- [ ] Update README

bcenker:
I started to implement this, but wanted to get your thoughts on the following before I continue:

1.  Do you think it would be better to have the user create an IAM role outside the scope of the VPC module and be forced to pass it in?  Most cases I've seen have a single flowlogs role at the account level that is used by every flowlog in the account.  Managing the IAM role in the VPC module could create an issue if the user wanted to destroy the VPC terraform that defined the role if that role was shared among other flowlogs in other VPC terraforms.  Additionally, if I have 30 (or some arbitrarily large number of) VPCs terraformed with this module and they all shared an IAM role that I created in the one of them, I wouldn't want to have to remember which one I defined the role in if I wanted to modify it's permissions.

[...]

PR #316, opened 2019-09-02
Add VPC Flow Logs capabilities

miguelaferreira:
# Description

This PR addresses issue #23 and adds the ability to enable VPC Flow Logs within the module.
The implementation tries to offer the user all the possibilities to configure Flow Logs while trying to keep the module interface as simple as possible. 

This PR extends the complete VPC example with the default behaviour (push logs to CloudWatch while having the module create all the necessary resources for that), and adds 4 new Flow Logs specific examples that illustrate how other configurations can be achieved. 

I made an implementation decision to keep the module interface simpler at the cost of having to run terraform twice (using `-target`) to make sure log destinations (or IAM role) exist before actually creating the VPC with Flow Logs enabled. The alternative would be to add at least two more boolean arguments to the module and then always specifying those in combination with providing actual log destination (or IAM role). I've kept that decision in a separate commit so it is easier to revert in case people find it better to have those extra boolean flags.


antonbabenko:
Hi, Miguel!

Thanks for this great pull-request!

There are just a few things which I'd like to change to make this more compatible with the rest. One of it is that there should be no need to run apply with `-target`. All examples should follow - init, plan, apply, destroy cycle.

Another one is that there is [terraform-aws-s3-bucket module](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket) which can be used as an example.

I will update the code and fix a couple of minor spelling mistakes during this week.

miguelaferreira:
Thanks for looking into this Anton. Please do let me know if I can help with the modifications you think should be made.

nicolaevladescu:
Hi, what's the hold up on this? Can i help with something?

miguelaferreira:
Nothing from my part @nicolaevladescu. I think @antonbabenko still wants to work this out better before merging.

antonbabenko:
Yes, I started looking into it and have been testing it in detail earlier this week. I think we will get it merged during next week.

antonbabenko:
Yes, I started reviewing and testing it back in October and almost finished it before switched to other things. I still have it in a local branch, and will resume once I switch to this repo.

babilen5:
I've just tried the changes in this PR and think that it would be great if they could get incorporated into the VPC module.

Thanks for your work on this, @miguelaferreira!

antonbabenko:
I continue working on this pull-request.

I now need this feature for my paying customer who has to pass [CIS - 2.9 – Ensure VPC flow logging is enabled in all VPCs](https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-cis-controls.html#securityhub-cis-controls-2.9).

antonbabenko:
🎉 [v2.26.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.26.0) has been just released with VPC Flow Log supported.

Thanks @miguelaferreira for your work. I have combined several examples related to the VPC flow log into one and removed dependency on S3 bucket module, so that bucket should always be created outside of this module if the user wants to push logs there.

Please let me know if there are any issues.

madddi:
@antonbabenko Updating to 2.26 results in an error during apply for me:

```
Error: Invalid index

  on .terraform/modules/default_vpc/terraform-aws-modules-terraform-aws-vpc-2ac8bf5/outputs.tf line 1008, in output "vpc_flow_log_id":
1008:   value       = aws_flow_log.this[0].id
    |----------------
    | aws_flow_log.this is empty tuple

The given key does not identify an element in this collection value.
```

antonbabenko:
Fixed in 8554bfd.

[v2.27.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.27.0) has been just released.

PR #326, opened 2019-09-09
Add subnets and tagging for EKS

chrisfowles:
# Adds functionality to make deploying EKS resources easier.

This change adds subnets for EKS worker nodes and configures tagging for vpc resources as per the aws requirements (https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html#vpc-tagging)


nehanikam:
@chrisfowles Quick question, the code looks fine but it seems that it was never merged in master. Was there any specific reason for this? Or is this still in the pipeline?

PR #337, opened 2019-10-01
fix: Return correct route table when enable_public_redshift is set

skorfmann:
# Description

In case the input variable `var.enable_public_redshift` is true, and no dedicated redshift route table is created, the output `redshift_route_table_ids` contains the `private` route table accidentally. It should contain the `public` route table though.


DrFaust92:
looks good to me, @bryantbiggs?

CC @antonbabenko

DrFaust92 (approved):

antonbabenko:
Thanks @skorfmann ! It took us almost 2 years 🥈 

[v3.5.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v3.5.0) has been just released.

PR #340, opened 2019-10-08
fix ipv6 enable

lifeofguenter:
fixes the following current issue:

```
Error: Error in function call

  on .terraform_stage/modules/main_vpc/terraform-aws-modules-terraform-aws-vpc-5358041/main.tf line 917, in resource "aws_route" "private_ipv6_egress":
 917:   route_table_id              = element(aws_route_table.private.*.id, count.index)
    |----------------
    | aws_route_table.private is empty tuple
    | count.index is 0

Call to function "element" failed: cannot use element function with an empty
list.
```

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -912,7 +912,7 @@ resource "aws_route" "private_nat_gateway" {
 }
 
 resource "aws_route" "private_ipv6_egress" {
-  count = var.enable_ipv6 ? length(var.private_subnets) : 0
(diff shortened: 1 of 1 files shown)


antonbabenko:
Great catch, @lifeofguenter !

[v2.21.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.21.0) has been just released.

PR #341, opened 2019-10-10
feat: Attach bucket policy for S3 endpoints

lliknart:
# Description

In case where s3 endpoints are enable, you could now attach a policy for this endpoint.
For exemple:

```
data "aws_iam_policy_document" "s3-access" {
  policy_id = "S3Access"
  version   = "2012-10-17"

  statement {
    sid    = "ReadAccess"
    effect = "Allow"
    principals {
      type        = "AWS"
      identifiers = ["*"]
    }
    actions = [
      "s3:GetObject",
    ]
    resources = [
      "arn:aws:s3:::my-bucket-in-read-only/*",
    ]
  }

  statement {
    sid    = "WriteAccess"
    effect = "Allow"
    principals {
      type        = "AWS"
      identifiers = ["*"]
    }
    actions = [
      "s3:PutObject",
    ]
    resources = [
      "arn:aws:s3:::my_first_bucket/*",
      "arn:aws:s3:::my_second_bucket/*",
    ]
  }
}
```

you could use the following lines to attach this policy to the S3 endpoint:

```
module "vpc_with_s3_endpoints" {
  source = "****"
  [...]
  enable_s3_endpoint = true
  s3_endpoint_policy = data.aws_iam_policy_document.s3-access.json
  [...]
```


patoarvizu:
Friendly `ping` on this... is there anything holding it back (other than the branch being out of date)?

variables.tf
```diff
@@ -316,6 +316,12 @@ variable "enable_s3_endpoint" {
   default     = false
 }
 
+variable "s3_endpoint_policy" {
+  description = "Provide a policy for the S3 endpoint"
+  type        = string
+  default     = ""
```
bryantbiggs, line 322: will this fail if users do not provide a policy - I think this should perhaps be `null` instead.

bryantbiggs (changes requested):
thank you for the PR @lliknart ! could you update the [`complete-vpc`](https://github.com/terraform-aws-modules/terraform-aws-vpc/tree/master/examples/complete-vpc) to reflect your proposed changes here and make sure the changes work successfully by provisioning the example please? we will need to make sure that the default behavior does not disrupt existing users, while allowing users the ability to provide a custom policy. thank you!

PR #343, opened 2019-10-17
feat: create route tables per AZ for intra subnets

iamnande:
#Description

This change implements a per-AZ route table for the intra subnets that get created as a part of the module.

It uses the same naming scheme to keep feature parity with the other subnets.


gillg:
Your MR seems to be review, you push lot of bad/not purpose "formating" changes.
Could you keep only changes related to routes tables per AZ ?

And what your issue motivating this change ? I don't undertand why you done this.

iamnande:
6 months later, not sure why some of the formatting (`terraform fmt`) were included and this clearly didn't receive the time initially. ☮️

PR #359, opened 2019-11-26
Enable EKS tagging

nehanikam:

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,18 @@ Since AWS Lambda functions allocate Elastic Network Interfaces in proportion to 
 
 You can add additional tags with `intra_subnet_tags` as with other subnet types.
 
+## EKS (Kubernetes) Resource Tagging and Subnets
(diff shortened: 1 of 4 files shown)


Grejeru:
Very nice addition, but what if there are multiple clusters on same VPC? It should include possibility to create multiple instances of `kubernetes.io/cluster/<cluster_name>` tags.

bryantbiggs:
thanks @nehanikam for this PR :tada: I don't know if this is something we will be incorporating into the module at this time. I think the best option for something like this is to have a flexible tagging scheme. This PR looks to make changes to a wide range of networking resources and I don't think that complexity is necessary at this time. There are a number of EKS deployments using this module as is so I think we will have to pass on this proposed change

DrFaust92:
FYI, subnet tagging is no longer needed

antonbabenko:
Closing this issue, as discussed above by @bryantbiggs and @DrFaust92

Issue #364, opened 2019-12-02
The name "for_each" is reserved for use in a future version of Terraform.


Noam-greenberrg:
Hi 
when using the last ver we are getting this error 

 996:   for_each = var.customer_gateways

The name "for_each" is reserved for use in a future version of Terraform.

Terraform v0.12.5
+ provider.aws v2.40.0
+ provider.random v2.2.1
+ provider.template v2.1.2

Your version of Terraform is out of date! The latest version
is 0.12.16. You can update by downloading from www.terraform.io/downloads.html

dpiddockcmp:
`for_each` was added in Terraform 0.12.6. Please update your Terraform

Noam-greenberrg:
HI 
My ver is ":0.12.16." so ...

dpiddockcmp:
Your original post says that you're using Terraform 0.12.5.

Noam-greenberrg:
yes you are correct
i will upgrade and test

xanderflood:
I am currently getting this same message with:

```
$ terraform version
Terraform v0.12.13
+ provider.template v2.1.2
```

I use for_each in several places in my config, but this error only appears when I attempt to use it with a module. Is for_each currently unsupported with modules?

antonbabenko:
`for_each` is a reserved name for an argument but it is used inside of a module itself.

Make sure that your wrapping module (if you make it) does not have `for_each` as an argument, empty cache folder `.terraform`.

janjagusch:
> I am currently getting this same message with:
> 
> ```
> $ terraform version
> Terraform v0.12.13
> + provider.template v2.1.2
> ```
> 
> I use for_each in several places in my config, but this error only appears when I attempt to use it with a module. Is for_each currently unsupported with modules?

This feature is not supported yet. More details [here](https://www.hashicorp.com/blog/hashicorp-terraform-0-12-preview-for-and-for-each/)

> For a long time, users have wished to be able to use the count meta-argument within module blocks, allowing multiple instances of the same module to be created more easily.
> 
>  Again, we have been laying the groundwork for this during Terraform 0.12 development and expect to complete this work in a later release. Along with count, module blocks will also accept the new for_each argument described for resources above, with similar results.
>
> This feature is particularly complicated to implement within Terraform's existing architecture, so some more work will certainly be required before we can support this. To avoid further breaking changes in later releases, 0.12 will reserve the module input variable names count and for_each in preparation for the completion of this feature.

[...]

Issue #361, opened 2019-11-27
The name "for_each" is reserved for use in a future version of Terraform


edgarcolque:
.terraform/modules/vpc/terraform-aws-modules-terraform-aws-vpc-d97bf1c/main.tf line 996, in resource "aws_customer_gateway" "this":
 996:   for_each = var.customer_gateways

The name "for_each" is reserved for use in a future version of Terraform.

antonbabenko:
You are using master branch or `v2` of this module which requires Terraform 0.12.

You need to use a correct version of the module as described here: https://github.com/terraform-aws-modules/terraform-aws-vpc#terraform-versions

johndesp:
We are using the master branch and TFE 12.2 and we are getting the same error. We don't get this error when running on TFE 12.21 . We can't upgrade to 12.21 yet.
Any ideas on how we can get around the following error?

The name "for_each" is reserved for use in a future version of Terraform.

dpiddockcmp:
`for_each` was added in TF 0.12.6. You must use 0.12.6 or later. Ideally much later as there were quite some interesting bugs in the early releases.

Currently only one resource currently uses `for_each`: `aws_customer_gateway.this`. Two options here:
- the resource was added in 2.20.0 of the module. Pin your module definition to version 2.19.0
- if you need other changes added after 2.20.0, you can fork the module and remove the `aws_customer_gateway` resource locally

PR #365, opened 2019-12-02
Help users by specifying minimum versions

dpiddockcmp:
Couple of Issues raised lately caused by users using outdated versions of Terraform (#364 #361).

Use TF 0.12's minimum version hints to help. Also add a test to CircleCI to help ensure these are kept up to date. `terraform validate` isn't 100% accurate now with lazy `if` evaluation but better than nothing.

Based on similar work done in the [terraform-aws-eks](https://github.com/terraform-aws-modules/terraform-aws-eks/) module.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 
 terraform: &terraform
   docker:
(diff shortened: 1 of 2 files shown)


bryantbiggs:
superseded by #598

PR #367, opened 2019-12-05
feat: Add VPC Endpoint Names

DrFaust92:
# Description

Add Name tag to all VPC Endpoints with default values
Add suffix variables to all VPC Endpoint names

+ Minor changes to Descriptions of some endpoint names

### DIFF
--- a/vpc-endpoints.tf
+++ b/vpc-endpoints.tf
@@ -13,10 +13,10 @@ resource "aws_vpc_endpoint" "s3" {
   vpc_id       = local.vpc_id
   service_name = data.aws_vpc_endpoint_service.s3[0].service_name
   tags = merge(
+    local.vpce_tags,
     {
-      Name = "${var.name}-${var.s3_vpc_endpoint_suffix}"
-    },
-    local.vpce_tags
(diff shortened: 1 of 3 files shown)


vpc-endpoints.tf
```diff
@@ -12,7 +12,12 @@ resource "aws_vpc_endpoint" "s3" {
 
   vpc_id       = local.vpc_id
   service_name = data.aws_vpc_endpoint_service.s3[0].service_name
-  tags         = local.vpce_tags
+  tags = merge(
+    local.vpce_tags,
+    {
+      "Name" = "${var.name}-${var.vpc_endpoint_suffix}"
```
antonbabenko, line 18: Why do you want to override `Name` tag like this?

You can already set `var.vpc_endpoint_tags = { Name = "my-vpcname-vpc-endpoint" }` to have the same effect, no?

DrFaust92: true, but this to align with other resources the add the name tag. default for this resources is that there is no name in the console.

i'll close this for now while i think of a better solution. ill open a different PR for the doc fixes

PR #368, opened 2019-12-05
Default value for booleans should be false so they do not show up as required in Terraform Registry

robin-norwood:
…required in Terraform registry

# Description

Before this change, several values show up as "required" that probably should not be:

Screenshot_2019-12-05 terraform-aws-modules vpc aws Terraform Registry


antonbabenko:
Hi @robin-norwood !

Thanks for opening this issue here but I think this problem has to be fixed in Terraform Registry and Terraform Cloud where `null` values are not respected properly. There is an open issue already - https://github.com/hashicorp/terraform/issues/21417

There is nothing that we should do in this module.

PR #370, opened 2019-12-05
Use Availability Zone IDs

### LINKED ISSUE #234: Feature Req: Support AZ ID and name on subnet creation
We are currently using this module for VPC creation. In order to support alignment across a given set of accounts for an upcoming project, we would like to have the ability to set the Availability Zone ID rather than the name when creating a subnet.

https://www.terraform.io/docs/providers/aws/r/subnet.html#argument-reference

Currently the module only supports name (availability_zone argument) when creating a subnet.

Regards,
Ryan

DrFaust92:
# Description

Use Availability IDs instead of names

Closes #234 & #378


DrFaust92:
@antonbabenko, sorry for the delay. i've got to testing this and seems ok to me.
I tested with the new added example and with the regular simple vpc example

antonbabenko:
Thanks @DrFaust92 for this PR!

I have simplified it a bit by removing the extra variable and updated an existing `simple-vpc` example.

[v2.25.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.25.0) has been just released.

PR #372, opened 2019-12-10
Add ability to define custom DB Subnet Group Name

stuzlogle:
# Description

TF Will destroy and recreate any aws_db_subnet_group resource which has a name change. Allowing the module to specify a name allows users to import existing resources into the vpc managed by the module.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -359,7 +359,7 @@ resource "aws_subnet" "database" {
 resource "aws_db_subnet_group" "database" {
   count = var.create_vpc && length(var.database_subnets) > 0 && var.create_database_subnet_group ? 1 : 0
 
-  name        = lower(var.database_subnet_group_name)
+  name        = var.database_subnet_group_name != null ? lower(var.database_subnet_group_name) : lower(var.name)
   description = "Database subnet group for ${var.name}"
   subnet_ids  = aws_subnet.database.*.id
 
--- a/main.tf
+++ b/main.tf
@@ -443,7 +443,7 @@ resource "aws_subnet" "elasticache" {
 resource "aws_elasticache_subnet_group" "elasticache" {
   count = var.create_vpc && length(var.elasticache_subnets) > 0 && var.create_elasticache_subnet_group ? 1 : 0
 
-  name        = var.name
+  name        = var.elasticache_subnet_group_name != null ? lower(var.elasticache_subnet_group_name) : lower(var.name)
   description = "ElastiCache subnet group for ${var.name}"
   subnet_ids  = aws_subnet.elasticache.*.id
 }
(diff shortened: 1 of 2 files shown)


y-batsianouski:
Upvote for this PR. It's really useful when you need to import existed resources into this module and you have over 50 RDS instances in existed db subnet group on production :)

variables.tf
```diff
@@ -10,6 +10,18 @@ variable "name" {
   default     = ""
 }
 
+variable "database_subnet_group_name" {
+  description = "Name to be used on DB Subnet Group resource as identifier"
```
y-batsianouski, line 14: It would be nice to add rules for the DB subnet group name.
`Must contain from 1 to 255 characters. Alphanumeric characters, spaces, hyphens, underscores, and periods are allowed.`

main.tf
```diff
@@ -443,7 +443,7 @@ resource "aws_subnet" "elasticache" {
 resource "aws_elasticache_subnet_group" "elasticache" {
   count = var.create_vpc && length(var.elasticache_subnets) > 0 && var.create_elasticache_subnet_group ? 1 : 0
 
-  name        = var.name
+  name        = var.elasticache_subnet_group_name != null ? lower(var.elasticache_subnet_group_name) : lower(var.name)
```
y-batsianouski, line 446: You added `lower(var.name)`. It may break infrastructure of the current users of this module during upgrade t the new version.
Not sure you need add `lower()` here at all.

y-batsianouski (changes requested):

main.tf
```diff
@@ -359,7 +359,7 @@ resource "aws_subnet" "database" {
 resource "aws_db_subnet_group" "database" {
   count = var.create_vpc && length(var.database_subnets) > 0 && var.create_database_subnet_group ? 1 : 0
 
-  name        = lower(var.name)
+  name        = var.database_subnet_group_name != null ? lower(var.database_subnet_group_name) : lower(var.name)
```
bryantbiggs, line 362: I think we should be able to get away with:
```suggestion
  name        = coalesce(lower(var.database_subnet_group_name, var.name))
```

main.tf
bryantbiggs, line 446: ya, not sure why some have `lower()` and some don't, but lets maintain status quo for now:
```suggestion
  name        = coalesce(var.elasticache_subnet_group_name, var.name)
```

variables.tf
bryantbiggs: agreed - I think this would be helpful for users to know if there is criteria that must be followed

bryantbiggs:
@stuzlogle left you some feedback - I know its been a minute since you opened the PR so let me know if you want us to carry it over the line

Issue #257, opened 2019-05-13
Default VPC Security Group Tags


e-moshaya:
Is there a way to set the vpc's default security group tags? I'm not referring to the default VPC security group but the default security group for the current VPC created through this module.

sirket:
You could simply add this to main.tf in the module:

resource "aws_default_security_group" "default" {
  vpc_id = "${aws_vpc.this.id}"

  ingress {
    protocol  = -1
    self      = true
    from_port = 0
    to_port   = 0
  }

  egress {
    from_port   = 0
    to_port     = 0
    protocol    = "-1"
    cidr_blocks = ["0.0.0.0/0"]
  }

   tags = "${merge(map("Name", format("%s", var.name)), var.tags, var.vpc_tags)}"
}

That said- the default VPC security group really should not be used for anything so leaving off the tags might make sense.

ndench:
I use the default security group for most things because it allows everything in the vpc to communicate. I only create other security groups when I want the outside world to communicate with my infrastructure (ie. ssh, web, etc).

Can confirm that the `aws_default_security_group` workaround from @sirket works well but it would be nice for the module to attach tags to the security group. If you're creating multiple VPC's in the same region, it quite hard to tell their default security groups apart unless they're tagged.

DrFaust92:
closed by #382

PR #382, opened 2020-01-12
feat: manage default security group

### LINKED ISSUE #256: Add option to clear default security group rules
The default security group is wide open to itself but this will show up during many security audits. As the default security group really shouldn't be used for anything- would it make sense to clear the rules either by default- or at least provide an option to do so?

e.g.

resource "aws_default_security_group" "default" {
   count = "${var.create_vpc && var.disable_default_sg ? 1 : 0}"

  vpc_id = "${aws_vpc.this.id}"
}

bryantbiggs:
# Description

- Add ability to manage default security group as well as ability to add/remove default security group rules

Tested with complete example setup. Default values provided in variables for ingress/egress rules match what is created by default with VPC, example shows how to remove both the ingress and egress rules and block all traffic. Note - when removing the default security group control from terraform the default security group won't be destroyed and the ingress/egress rules at that time will be left as is and the resource will no longer be controlled by terraform.

Closes #256 #257


main.tf
```diff
@@ -54,6 +54,48 @@ resource "aws_vpc_ipv4_cidr_block_association" "this" {
   cidr_block = element(var.secondary_cidr_blocks, count.index)
 }
 
+resource "aws_default_security_group" "this" {
+  count = var.create_vpc && var.manage_default_security_group ? 1 : 0
+
+  vpc_id = element(concat(aws_vpc.this.*.id, [""]), 0)
```
antonbabenko, line 60: `vpc_id = aws_vpc.this[0].id` shoudl be enough since there is `count` already.

main.tf
```diff
@@ -54,6 +54,48 @@ resource "aws_vpc_ipv4_cidr_block_association" "this" {
...
+      description      = lookup(egress.value, "description", null)
+      from_port        = lookup(egress.value, "from_port", 0)
+      to_port          = lookup(egress.value, "to_port", 0)
+      protocol         = lookup(egress.value, "protocol", "-1")
+    }
+  }
+
+  tags = merge(
+    {
+      "Name" = format("%s", var.default_security_group_name)
```
DrFaust92, line 92: if the default SG is set to be managed and the `default_security_group_name` left with no change to default value it creates a an empty `Name` tag. its not a problem functionally but just FYI if you want to change this.

variables.tf
```diff
@@ -2259,12 +2259,58 @@ variable "elasticache_outbound_acl_rules" {
...
+  description = "Name to be used on the default security group"
+  type        = string
+  default     = ""
+}
+
+variable "default_security_group_ingress" {
+  description = "List of maps of ingress rules to set on the default security group"
+  type        = list(map(string))
+
+  default = [
```
antonbabenko, line 2278: Maybe defaults should be empty, because "When Terraform first adopts the Default Security Group, it immediately removes all ingress and egress rules in the Security Group." (from https://www.terraform.io/docs/providers/aws/r/default_security_group.html )

And, if I understand correctly, Terraform won't put original values back, if the user switches off `manage_default_security_group` afterwards.

Previously, there were attempts to manage this resource in this module, but it was a bit too advanced for some use-cases and Terraform 0.11 which we had back then.

main.tf
bryantbiggs: good point - maybe I add a check for a provided name value otherwise provide one like "default"

variables.tf
bryantbiggs: good point - I think you're right, might be better to use `null` here as it seems to be the safest route usually

main.tf
bryantbiggs: ended up adding a default value of `default` for the variable

antonbabenko:
Thanks, @bryantbiggs !

[v2.44.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.44.0) has been released.

PR #418, opened 2020-03-20
Added support for default security group for sqs endpoint.

exocom:
## Description

1. Added flag to use default security group for sqs endpoint.

2. Added support for only adding 1 subnet per availability zone. Endpoints are limited to 1 subnet per availability zone.

```tf
  azs = ["us-west-2a", "us-west-2b", "us-west-2c"]
  private_subnets = [
    cidrsubnet(local.cidr, 6, 0), cidrsubnet(local.cidr, 6, 4), cidrsubnet(local.cidr, 6, 8), 
    cidrsubnet(local.cidr, 8, 4), cidrsubnet(local.cidr, 8, 20), cidrsubnet(local.cidr, 8, 36), 
  ]

// Will use subents:
// cidrsubnet(local.cidr, 6, 0), cidrsubnet(local.cidr, 6, 4), cidrsubnet(local.cidr, 6, 8)
```

3. Limited availability zones to take only first 3. This is because SQS endpoint is only supported in a,b,c availability zones. Sadly d+ do not. (Tested in us-west-2)
 
NOTE: This requires that AZ-ZONES be in order.
```tf
Works in order:
azs = ["us-west-2a", "us-west-2b", "us-west-2c", "us-west-2d"]
 
Will not work
azs = ["us-west-2d", "us-west-2b", "us-west-2a", "us-west-2a"]
```

## Motivation and Context

We use terragrunt and it is not possible to reference an external security group because it causes a circular dependency. Based on the example and AWS documentation using the default security group is the norm.

## Breaking Changes

No

## How Has This Been Tested?

This was test by using multiple input configurations of AZ + subnets.

### DIFF
--- a/examples/complete-vpc/main.tf
+++ b/examples/complete-vpc/main.tf
@@ -2,11 +2,6 @@ provider "aws" {
   region = "eu-west-1"
 }
 
-data "aws_security_group" "default" {
-  name   = "default"
-  vpc_id = module.vpc.vpc_id
-}
-
 module "vpc" {
   source = "../../"
 
@@ -104,7 +99,7 @@ module "vpc" {
   # VPC endpoint for SQS
   enable_sqs_endpoint              = true
   sqs_endpoint_private_dns_enabled = true
-  sqs_endpoint_security_group_ids  = [data.aws_security_group.default.id]
+  sqs_endpoint_default_security_group  = true
 
   # VPC Flow Logs (Cloudwatch log group and IAM role will be created)
   enable_flow_log                      = true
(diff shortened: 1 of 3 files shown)


bryantbiggs:
hi @exocom - thank you for the PR but I don't think we will be making this change unfortunately. it would make SQS endpoints different from the rest of the interface endpoints, and its not quite idiomatic just because Terragrunt does not support this way of providing the security group as shown in the example. its also a good practice to not rely on the default security group and instead create custom security groups and use them as desired for the VPC endpoints

Issue #126, opened 2018-05-16
create_database_subnet_group does not work


GeekOnGadgets:
Hi, 

I created the VPC using this module and according to documentation `create_database_subnet_group` is enabled by default. Now when I try to create RDS and try to link to VPC it fails and gets create to the default VPC. The below how i achieve this

```
# NETWORKING #
module "vpc" {
  source = "terraform-aws-modules/vpc/aws"
  version = "1.30.0"
  name = "my-vpc"

  cidr = "10.0.0.0/16"
  azs             = ["eu-west-1a", "eu-west-1b", "eu-west-1c"]
  private_subnets = ["10.0.1.0/24", "10.0.2.0/24"]
  public_subnets  = ["10.0.101.0/24", "10.0.102.0/24"]

  enable_nat_gateway = false
  create_database_subnet_group = true

  tags = {
    Terraform = "true"
    Environment = "dev"
  }
}
```

```
resource "aws_db_instance" "dev" {
  identifier = "dev"
  allocated_storage = 20
  storage_type = "gp2"
  engine = "postgres"
  engine_version = "9.6.6"
  instance_class = "db.t2.micro"
  name = "app"
  username = "abc"
  password = "abc"
  parameter_group_name = "default.postgres9.6"
  backup_retention_period = 7
  publicly_accessible = true
  skip_final_snapshot = true
  vpc_security_group_ids = ["${module.vpc.default_security_group_id}"]
  db_subnet_group_name = "${module.vpc.database_subnet_group}"
  tags {
    Terraform = "true"
    Environment = "dev"
  }
}
```

But I keep on getting this error  `Error creating DB Instance: InvalidParameterCombination: The DB instance and EC2 security group are in different VPCs`

Any help is appreciated. 

Thanks

antonbabenko:
You need to specify CIDR blocks in `database_subnets` as in [this example](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/55b5991c2b89d41639c4cad960cea42826c4743a/examples/complete-vpc/main.tf#L15)

PR #424, opened 2020-04-01
docs: Document create_database_subnet_group requiring database_subnets

acdha:
## Description

I ran into the same issue as https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/126 because there's nothing in the documentation making it clear that `create_database_subnet_group` only has an effect if you also specify `database_subnets`.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -209,7 +209,7 @@ variable "create_elasticache_subnet_route_table" {
 }
 
 variable "create_database_subnet_group" {
-  description = "Controls if database subnet group should be created"
+  description = "Controls if database subnet group should be created (n.b. database_subnets must also be set)"
   type        = bool
   default     = true
 }


antonbabenko:
Thank you, @acdha !

acdha:
Thank you for such a handy module!

ahummel25:
@antonbabenko @acdha Any idea is if the new db subnet group name that gets created can be an output from the vpc module? I need this as an input to an `rds` instance. Not seeing any docs on output availability from vpc module.

antonbabenko:
https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/23b0a02d2a3225a05ddf5f31d9a32728bf5505c3/outputs.tf#L131-L134

PR #426, opened 2020-04-03
feat: add az tags on subnets

vfoucault:
* in order to best find network resources, the `az` and `az_id` are added to subnet_resources

## Description

added three tags for the `aws_subnet` resource

* `az`: the short letter for the specified availability zone

* `az_id`: the availability zone id 

* `serial`: the subnet number with `count`
## Motivation and Context

In order to be able to easy pinpoint a subnet, I have the option to build a complex name parser on each and every component that will have to find vpcs and subnets, or to update the `vpc module` to add tags that can't be added elsewhere.

Using tags for subnet resources I can filter subnets with criteria such as the availability zone letter, the AWS AZID, or even the serial

## Breaking Changes

No breaking changes. Only new tags. The sole issue would be that one uses `var.tags` with `az`, `az_id` or `serial`, then in this case it would be overwritten by these changes.

## How Has This Been Tested?

in `test_vpc.rb` we check the presence for the following tags:

* `az` with the proper value that can be extracted in the tests

* `az_id` presence only. as AZIDs are per AWS account

* `serial` present only, as we only test one subnet that could be either one or another


gillg:
Force tags for everyone seems not the best thing I think, but I like your idea to add azs indications on subnets.
Maybe reviewers will be ok with that.

bryantbiggs:
thanks @vfoucault for the PR - could you perhaps elaborate on your motivations or what you hope to solve if this PR was accepted? I feel like this is better suited to use the data source https://www.terraform.io/docs/providers/aws/d/subnet.html instead for this but want to see what problem you are solving for first

vfoucault:
@bryantbiggs the main goal is to easily find subnet(s) matching a specific az. 

Using the vpc module there is no possibility to do such thing. I may create subnets but no way to pinpoint a specific az.

This PR allows subnets to have the availability zone tag and also the availability zone ID. Finally a serial giving an number for a subnet in a vpc.

bryantbiggs:
thanks for the info @vfoucault, however as I mentioned before, can't you achieve this through the subnet data source?

```hcl
data "aws_subnet" "private_a" {
  vpc_id = module.vpc.vpc_id
  filter {
    name   = "tag:Name"
    values = ["main-private-us-east-1a"]
  }
}

data "aws_subnet" "private_b" {
  vpc_id = module.vpc.vpc_id
  filter {
    name   = "tag:Name"
    values = ["main-private-us-east-1b"]
  }
}

data "aws_subnet" "private_c" {
  vpc_id = module.vpc.vpc_id
  filter {
    name   = "tag:Name"
    values = ["main-private-us-east-1c"]
  }
}

locals {
  private_subnet_ids = [data.aws_subnet_private_a.id, data.aws_subnet_private_b.id, data.aws_subnet_private_c.id]
```

You can get all attributes of a subnet from this approach as well - [Ref](https://www.terraform.io/docs/providers/aws/d/subnet.html). I'm not saying your proposal is wrong or not worth merit, but it has quite strong assumptions for the community at large that uses the module when there are alternate methods for getting this information without everyone opting in to the changes.

vfoucault:
Thanks for the reply. You are correct, I could fetch subnets using `aws_subnet_ids`, using `aws_subnet` it would fail in case of having more than one subnet per az.

Still, if I want to filter subnets by AZ_ID it becomes a nightmare. I may only add `availability_zone_id` to `aws_subnet` and not to `aws_subnet_ids`

This means that I need to:

1) find all subnets given a composed named for all AZ
2) for each subnet using the `aws_subnet` data
3) filter out subnet that matches the proper AZ_ID

instead of filtering per AZ_ID

The other approached that I've looked at is to update the module to accept per subnet tags as referenced may times, but it is not the best approach as finding AZ_ID would have to be done outside of the module itself.

vfoucault:
Hi @bryantbiggs . Does it worth to rebase on master ? Is there any chance for this to be merged ? Tia

vfoucault:
Hey @bryantbiggs, @antonbabenko  it's been quite some time. Would you mind check this PR again ? Thanks in advance

bryantbiggs:
I'm sorry, but I don't think this is something we want adopt in this module

antonbabenko:
Hi @vfoucault !

Sorry for not coming back to this PR earlier.

I agree with Bryant that it is already possible to achieve using a combination of data sources and I don't think we can have this functionality in the module because not everyone needs the features introduced here.

For now, let's close this issue.

PR #438, opened 2020-04-27
docs: Add endpoint example for Cloudwatch

cob16:
## Description
Add endpoint example for Cloudwatch to `complete-vpc/main.tf`

## Motivation and Context
The complete VPC example does not include how to allow Cloudwatch endpoints

## Breaking Changes
None

## How Has This Been Tested?
- Vars used in the project to good success
- `terraform validate` passes


antonbabenko:
Thanks for the PR, @cob16 . I don't think there is a need to put cloudwatch into example code because there are already other similar types of VPC endpoints there.

PR #440, opened 2020-05-04
feat: Added external gateway for intra subnets

vfabrizi:
## Description
Add create_intra_external_gateway_route and intra_external_gateway options

## Motivation and Context
We are using your vpc modules, and we need to be able to specify a default gateway for intra subnets. The (unfortunate) use case is that we are supporting an AWS Workspace setup where the traffic has to go through a Sophos gateway. Hence the need to be able to use that LB as the default gateway.

## Breaking Changes
No

## How Has This Been Tested?
This was test by using multiple input configurations with intra subnet relevance.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -277,6 +277,18 @@ resource "aws_route_table" "intra" {
   )
 }
 
+resource "aws_route" "intra_external_gateway" {
+  count = var.create_vpc && length(var.intra_subnets) > 0 && var.intra_external_gateway != "" && var.create_intra_external_gateway_route ? 1 : 0
+
+  route_table_id         = aws_route_table.intra[0].id
+  destination_cidr_block = "0.0.0.0/0"
+  gateway_id             = var.intra_external_gateway
+
+  timeouts {
+    create = "5m"
+  }
+}
+
 ################
 # Public subnet
 ################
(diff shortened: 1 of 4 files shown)


outputs.tf
```diff
@@ -198,6 +198,16 @@ output "intra_subnets_ipv6_cidr_blocks" {
   value       = aws_subnet.intra.*.ipv6_cidr_block
 }
 
+output "create_intra_external_gateway_route" {
```
DrFaust92, line 201: not sure it makes sense to output this

outputs.tf
```diff
@@ -198,6 +198,16 @@ output "intra_subnets_ipv6_cidr_blocks" {
...
+  description = "Controls if an external gateway route for intra subnets should be created"
+  value       = var.create_intra_external_gateway_route
+}
+
+output "intra_external_gateway" {
+  description = "An external gateway for the intra subnets"
+  value       = var.intra_external_gateway
```
DrFaust92, line 208: same, these are already given as a var

main.tf
```diff
@@ -277,6 +277,18 @@ resource "aws_route_table" "intra" {
   )
 }
 
+resource "aws_route" "intra_external_gateway" {
+  count = var.create_vpc && length(var.intra_subnets) > 0 && var.intra_external_gateway != "" && var.create_intra_external_gateway_route ? 1 : 0
+
+  route_table_id         = aws_route_table.intra[0].id
+  destination_cidr_block = "0.0.0.0/0"
```
DrFaust92, line 284: maybe add this as a var with "0.0.0.0/0" default value

DrFaust92 (changes requested):

Issue #332, opened 2019-09-23
create default vpc security group for vpc interface endpoints


aavileli:
A default security group should be created for the vpc and associated to interface vpc endpoints if one is not supplied.

antonbabenko:
Hi Ashley,

This is not always a good idea to assume anything and create resources on behalf of a user. This module requires you to create a security group and pass it as an argument. More transparency is better.

Closing this issue.

aavileli:
Absolutely agree but when using an upstream module with terragrunt . This will create a circular dependency as the VPC needs to be created first then an SG needs to be added to the module which requires the VPC id.
Therefore asked if a parameter could be set which would create a default SG to be attached to the VPC endpoint. The module encapsulates creation of VPC and endpoints.

antonbabenko:
You are right, I didn't think about this use-case though I felt that creation of VPC endpoints _should_ also be available outside of this module (separate `terraform-aws-module`, for eg).

The flow to do this now is by calling the same VPC module with different arguments:
VPC (without VPC endpoints) => SG => VPC (with VPC endpoints)

If we create a temporary SG together with the VPC and use it for VPC endpoints it will be fine for the first run, but after that user will still prefer to set another security group and get rid of temporary SG - which will still result in a two-step execution.

Do you have ideas on how to solve this and have a single step execution?

aavileli:
I think creation of VPC endpoints should be out of this module. But if it encapsulates creation of the VPC and endpoints, then there could be a boolean flag for vpc endpoints use default security group for vpc endpoints. 

Then in the vpc-endpoints.tf this could be added

data "aws_security_group" "default" {
  name   = "default"
  vpc_id = aws_vpc.id
}

security_group_ids = var.use_default_security_group == true ? data.aws_security_group.default.vpc_id : var.ssm_endpoint_security_group_ids 

To fix my issue with circular dependency in terragrunt I have now created a quirk fix by encapsulated your module into another module where I can create SG dependency to your module.

antonbabenko:
I think until VPC endpoints are not externalized into a separate module, there will be a need in a quirk fix as you described. Maybe you can open a PR and describe in README your solution? It would definitely help some people.

tomalok:
I may be in a similar dependency loop, but with regards to `..._endpoint_subnet_ids`.

[...]

PR #454, opened 2020-06-05
feat: Add ability to supply vpc_id along with endpoint creation

polymorcodeus:
## Description

Adding functionality to use existing VPC to create endpoints. Added two new variables, vpc_id and use_existing_vpc_id_for_endpoints, as well as adding conditional creation for the endpoints based on their input. If both of the new variables are ignored/left to default, no existing module functionality should be affected.

## Motivation and Context

Allows module to be used with an existing VPC to create endpoints not using the default security group. Also solves the circular reasoning issue below, allowing creation of a VPC, then the endpoint security group, then the endpoints.

https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/332

## Breaking Changes

Not that I've seen.

## How Has This Been Tested?

Tested example code, providing two ids listed in variables doc.

Unable to get test-kitchen working on my mac to run the tests in the repo. Apologies.


polymorcodeus:
Excluded non-subnet dependent endpoints from proposed changes. Will be updating PR shortly.

PR #455, opened 2020-06-05
feat: Enable support for Terraform 0.13 as a valid version by setting minimum version required

### LINKED ISSUE #453: Support terraform 0.13
When running the [0.13 beta](https://releases.hashicorp.com/terraform/0.13.0-beta1/), I get the following error:

```
Error: Unsupported Terraform Core version

  on .terraform/modules/client_vpc/terraform-aws-vpc-2.38.0/versions.tf line 2, in terraform:
   2:   required_version = "~> 0.12.6"

Module module.client_vpc (from terraform-aws-modules/vpc/aws) does not support
Terraform version 0.13.0-beta1. To proceed, either choose another supported
Terraform version or update this version constraint. Version constraints are
normally set for good reason, so updating the constraint may lead to other
errors or unexpected behavior.
```

bryantbiggs:
## Description
- Set minimum version required which will allow for 0.13 to start being used with module

## Motivation and Context
Closes #453

### DIFF
--- a/versions.tf
+++ b/versions.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = "~> 0.12.6"
+  required_version = ">= 0.12.6"
 
   required_providers {
     aws = "~> 2.53"
--- a/versions.tf
+++ b/versions.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 0.12.6"
+  required_version = ">= 0.12.6, < 0.14"
 
   required_providers {
     aws = "~> 2.53"


versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = "~> 0.12.6"
```
antonbabenko, line 2: This is too optimistic that things won't break in 0.14 :)

Instead, can you write something like `~> 0.12.6 || ~> 0.13.0` (I didn't test this) ?

bryantbiggs: good point. I can add an upper-bounds to allow up to but not including 0.14 (similiar to what you wrote) https://www.terraform.io/docs/configuration/terraform.html#specifying-a-required-terraform-version

antonbabenko: I am thinking about this - `~> 0.12.6, ~> 0.13.0`. The point is that in 0.13.x may be something wrong and we may want to add another _excluding_ range after a comma.

bryantbiggs: ok give me some time to do a quick test to see if that style works, I can't tell for sure from just reading the docs. I'll get back to you

bryantbiggs: ok so the syntax you've proposed is not valid - the closest equivalent is what I have proposed in this PR

antonbabenko: Ok, good to know. Merging.

antonbabenko:
Thanks, Bryant!

[v2.39.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.39.0) has been just released.

PR #458, opened 2020-06-20
fix: Reorder tags to allow overriding Name tag in route tables

### LINKED ISSUE #406: database_route_table_tags: Name doesn't overwrite default
Using following in module "vpc" does not overwrite calculated Name tag:

  database_route_table_tags = {
    Name = "rt-db-simple-example"
  }

This works fine for public and private route tables but not for database route table. 
Believe it is the order of the maps in the merge on lines 184-191 in main.tf.  var.database_route_table_tags should be last allowing users wishes to rule.

DrFaust92:
## Description

reorder tags to allow overriding Name tag

## Motivation and Context

Closes #406

## Breaking Changes

No

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -190,11 +190,11 @@ resource "aws_route_table" "database" {
   vpc_id = local.vpc_id
 
   tags = merge(
-    var.tags,
-    var.database_route_table_tags,
     {
       "Name" = "${var.name}-${var.database_subnet_suffix}"
     },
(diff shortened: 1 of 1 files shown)


antonbabenko:
Thanks @DrFaust92 !

[v2.42.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.42.0) has been just released.

PR #459, opened 2020-06-22
feat: add database_subnet_group_name to outputs

colinmcdonald22:
## Description
This will add `database_subnet_group_name` to outputs.

## Motivation and Context
This output will be able to be passed directly into `aws_db_instance`'s `db_subnet_group_name` argument. Previously, one had to use something like `module.vpc.name`. Although this is (functionally) the same thing, relying on the `name` variable is implicit. Additionally, this makes DB subnet groups consistent with Elasticache subnet groups.

As far as I'm aware, there is no open issue for this.

## Breaking Changes
No breaking changes.

## How Has This Been Tested?
I've tested running the modified version locally and it produces output as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -627,6 +627,7 @@ It is possible to integrate this VPC module with [terraform-aws-transit-gateway 
 | database\_route\_table\_ids | List of IDs of database route tables |
 | database\_subnet\_arns | List of ARNs of database subnets |
 | database\_subnet\_group | ID of database subnet group |
+| database\_subnet\_group\_name | Name of database subnet group |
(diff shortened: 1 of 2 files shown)


dpiddockcmp:
Use the output [`database_subnet_group`](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/v2.44.0/outputs.tf#L136)

From the [Terraform docs](https://www.terraform.io/docs/providers/aws/r/db_subnet_group.html#id)
> `id` - The db subnet group name.

PR #474, opened 2020-07-28
fix: bumping terraform version from 0.12.6 to 0.12.7 in circleci to include regexall function

jonaskint:
## Description
Circleci tests are failing the last releases. It is complaining about the regexall function that is not included in the current 0.12.6 version.

## Motivation and Context
I saw that the regexall function was not recognized because of an older terraform version that is used in circlci tests. I bumped the terraform version in order to fix the test.

## Breaking Changes
None as far as I know.

## How Has This Been Tested?
I ran the oneliners from the circleci config file.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 
 terraform: &terraform
   docker:
-    - image: hashicorp/terraform:0.12.6
+    - image: hashicorp/terraform:0.12.29
   working_directory: /tmp/workspace/terraform
 
 jobs:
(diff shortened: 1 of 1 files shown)


jonaskint:
> Looks good to me.
> 
> A thought to maybe lower the version to the first version where the function is available instead? im guessing most people are not using the latest version and there might some changes there that might miss some other bugs. But its fine like this as well.

it was included in 0.12.7, i could lower the version number to that

DrFaust92:
@antonbabenko small change, can we merge?

PR #475, opened 2020-07-28
fix: Name precedence in tag merges

gregmoy:
## Description
The tag merge was eating the Name tag

## Motivation and Context
Per https://www.terraform.io/docs/configuration/functions/merge.html the later argument takes precedence
suffix variables were not changing names

## Breaking Changes
Will rename everything if you had them named the old way obviously

## How Has This Been Tested?
Ran it locally

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -42,7 +42,7 @@ resource "aws_vpc" "this" {
     var.vpc_tags,
     {
       "Name" = format("%s", var.name)
-    }
+    },
   )
 }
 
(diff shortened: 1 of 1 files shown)


DrFaust92:
some folks would like to be able to override the name tag, so i'm not sure this is something we would like to support.
in your PR's case Name cannot be overridden.

gregmoy:
@DrFaust92 can the documentation be updated to reflect this? Currently theres nothing that suggests theres a conflict between the `name` and `tags` options nor any indication that a `Name` entry in the tags would render the `_suffix` variables useless.

antonbabenko:
Closing this PR, as @DrFaust92 said. If anyone feels that documentation can be improved please open a dedicated PR.

Issue #478, opened 2020-08-01
AWS 3.0 provider


vladgh:
Hello, can you please update this module to support the new AWS ~> 3.0 provider?

Thank you in advance.

PR #479, opened 2020-08-03
feat: bump version of aws provider version to support 3.*

michalschott:
## Description
Bumping aws-provider requirement to ">= 2.57"

Should fix https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/478

## Motivation and Context
There is newer provider version available.

## Breaking Changes
As stated [here](https://github.com/terraform-providers/terraform-provider-aws/releases/tag/v3.0.0)

## How Has This Been Tested?
Ran it locally

### DIFF
--- a/versions.tf
+++ b/versions.tf
@@ -2,6 +2,6 @@ terraform {
   required_version = ">= 0.12.6, < 0.14"
 
   required_providers {
-    aws = "~> 2.57"
+    aws = ">= 2.57"
   }
 }
--- a/versions.tf
+++ b/versions.tf
@@ -1,7 +1,7 @@
 terraform {
-  required_version = ">= 0.12.6, < 0.14"
+  required_version = ">= 0.12.7, < 0.14"
 
   required_providers {
-    aws = ">= 2.57"
+    aws = ">= 2.57, < 4.0"
   }
 }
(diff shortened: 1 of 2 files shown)


versions.tf
```diff
@@ -2,6 +2,6 @@ terraform {
   required_version = ">= 0.12.6, < 0.14"
 
   required_providers {
-    aws = "~> 2.57"
+    aws = "~> 3.0"
```
dpiddockcmp, line 5: This is a really bad way to achieve what you're trying to do. Now the module is not compatible with v2 of the provider, which is a breaking change, and requires a major version bump. For this single line change.

The [Terraform documentation](https://www.terraform.io/docs/configuration/modules.html#provider-version-constraints-in-modules) recommends the use of `>=` in `required_providers` for exactly this reason. No release would have been necessary to support the new provider.

A better solution would be:
```suggestion
    aws = ">= 2.57"
```
Would only need a minor version release of the module.

Alternatively, if Anton wants to stay guarded about the AWS provider version:
```hcl
  aws = ">= 2.57, < 4.0"
```

michalschott: You're totally right 🤦

marcincuber:
Can we get this merged and released asap? It is not a breaking change and it will allow to start testing it with other modules which are already using v3 aws provider.

antonbabenko:
Thanks to all for the patience, [v2.45.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.45.0) has been just released.

PR #486, opened 2020-08-11
feat: Added support for terraform v0.13 and aws provider v3.0

herrLierb:
## Description
Terraform v0.13 feature `count` on modules, does make the variable `create_eks` obsolete.
Additionally aws provider v3.0 is released (no changes needed there as I see it)

## Motivation and Context
My project uses terraform 0.13 and this is one of the incompatible modules, I use.

## Breaking Changes
Variable `create_eks` is not needed anymore and may trigger warnings for users if they try to pass that variable to the module.

## How Has This Been Tested?
I cloned the code locally and created some VPCs with different configurations (with / without NAT GWs ect.)
Create, update and destroy worked as I expected.

It's my first PR though and I'm not sure how it should have been tested more thoroughly but I'm happy to learn!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 
 terraform: &terraform
   docker:
-    - image: hashicorp/terraform:0.12.6
+    - image: hashicorp/terraform:0.13.0
   working_directory: /tmp/workspace/terraform
 
 jobs:
--- a/README.md
+++ b/README.md
@@ -163,14 +163,14 @@ VPC Flow Log allows to capture IP traffic for a specific network interface (ENI)
 
 ## Conditional creation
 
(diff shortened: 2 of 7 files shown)


antonbabenko:
Hi @herrLierb !

This PR does more than just bumping required versions which have been just released in [v2.45.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.45.0), so I am going to close this PR now.

herrLierb:
Hi @antonbabenko 

sure, thanks for your comment and the release. The new release does fix my issue, though the `var.create_vpc` still is obsolete in future versions. I just thought it nice to clear that up right away.

Cheers!

antonbabenko:
I agree in general if there would be just a few users who rely on this module. In reality, every change to this module has to be implemented with BC compatibility in mind.

PR #487, opened 2020-08-17
fix: 0.13 upgrade

thomaspicquet:
## Description
Updated version so it loads the provider correctly

## Motivation and Context
Getting this error when running on Terraform cloud
```
Failed to instantiate provider "registry.terraform.io/-/aws" to obtain schema:
unknown provider "registry.terraform.io/-/aws"
```

## Breaking Changes
none

## How Has This Been Tested?
testing it locally

### DIFF
--- a/versions.tf
+++ b/versions.tf
@@ -3,7 +3,7 @@ terraform {
 
   required_providers {
     aws = {
-      source = "hashicorp/aws"
+      source  = "hashicorp/aws"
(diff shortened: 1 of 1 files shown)


antonbabenko:
The short syntax is fine here because this module works with 0.12, too.

See this for more info - https://www.terraform.io/docs/configuration/provider-requirements.html#v0-12-compatible-provider-requirements

PR #495, opened 2020-09-02
Remove general Name tag on examples

gugaiz:
## Description
This PR remove the global Name tag from some examples. This global tag overwrite all names (subnets mainly) with a common name that does not allow to identify resources later

## Motivation and Context
I don't think it is a good practice to overwrite all names, also for starters like myself, it is hard to find the reason when you are learning from the example provided

## Breaking Changes
No, this does not break things

## How Has This Been Tested?
I run the examples with that general name tag and all my subnets were properly named

### DIFF
--- a/examples/complete-vpc/main.tf
+++ b/examples/complete-vpc/main.tf
@@ -120,7 +120,6 @@ module "vpc" {
   tags = {
     Owner       = "user"
     Environment = "staging"
-    Name        = "complete"
   }
 
   vpc_endpoint_tags = {


antonbabenko:
`Name` tag is there to show that if a more concrete tag per resource type is not set (eg, `vpc_endpoint_tags`), it will use the default one provided.

The example is not wrong, and this is the desired experience.

Closing this one.

gugaiz:
@antonbabenko sorry, but it does not use the default one.. the default one for the public subnet will be `dev-vpc-public-us-east-2a` instead of `complete` so with that example you end up with all subnets called `complete` which I don't think it is very useful when you try to recognize them on the AWS console.

PR #497, opened 2020-09-09
feat: Add ability to pass custom IAM policy to VPC endpoints

### LINKED ISSUE #437: VPC Endpoint policy
I notice the module doesnt create a [policy](https://www.terraform.io/docs/providers/aws/r/vpc_endpoint.html#policy) for an endpoint. Is this something that will be added in a future release, or can be added?

kostyaplis:
## Description
In order to fix #437 
Inspired by #341 

Added ability to pass custom IAM policy as JSON string to any VPC endpoint that supports it atm

## Motivation and Context
Default `Full Access` endpoints policy is a security concern. At least, accordingly to ISO27001

## Breaking Changes
No breaking changes

Been tested with custom and default IAM policies.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -2372,3 +2372,207 @@ variable "create_egress_only_igw" {
   type        = bool
   default     = true
 }
+
+variable "s3_endpoint_policy" {
+  description = "Custom IAM policy for S3 VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "dynamodb_endpoint_policy" {
+  description = "Custom IAM policy for DynamoDB VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "codebuild_endpoint_policy" {
+  description = "Custom IAM policy for Codebuild VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "codecommit_endpoint_policy" {
+  description = "Custom IAM policy for Code Commit VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "sqs_endpoint_policy" {
+  description = "Custom IAM policy for SQS VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "secretsmanager_endpoint_policy" {
+  description = "Custom IAM policy for Secrets Manager VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "ec2_endpoint_policy" {
+  description = "Custom IAM policy for EC2 VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "ec2_autoscaling_endpoint_policy" {
+  description = "Custom IAM policy for EC2 Autoscaling VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "ecr_api_endpoint_policy" {
+  description = "Custom IAM policy for ECR API VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "ecr_dkr_endpoint_policy" {
+  description = "Custom IAM policy for ECR DKR VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "apigw_endpoint_policy" {
+  description = "Custom IAM policy for API Gateway VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "kms_endpoint_policy" {
+  description = "Custom IAM policy for KMS VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "sns_endpoint_policy" {
+  description = "Custom IAM policy for SNS VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "monitoring_endpoint_policy" {
+  description = "Custom IAM policy for CloudWatch Monitoring VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "logs_endpoint_policy" {
+  description = "Custom IAM policy for CloudWatch Logs VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "events_endpoint_policy" {
+  description = "Custom IAM policy for CloudWatch Events VPC endpoint"
+  type        = string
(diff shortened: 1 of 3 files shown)


variables.tf
```diff
@@ -2372,3 +2372,207 @@ variable "create_egress_only_igw" {
...
+  default     = ""
+}
+
+variable "ecr_api_endpoint_policy" {
+  description = "Custom IAM policy for ECR API VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "ecr_dkr_endpoint_policy" {
```
bryantbiggs, line 2430: likewise, I see ECR supported but I don't know if that includes the Docker registry API endpoint https://docs.aws.amazon.com/vpc/latest/userguide/integrated-services-vpce-list.html

vpc-endpoints.tf
```diff
@@ -1,3 +1,17 @@
+data "aws_iam_policy_document" "default" {
```
bryantbiggs, line 1: we shouldn't need this, instead we can set the policy to `null` most likely. see below

vpc-endpoints.tf
```diff
@@ -12,6 +26,7 @@ resource "aws_vpc_endpoint" "s3" {
 
   vpc_id       = local.vpc_id
   service_name = data.aws_vpc_endpoint_service.s3[0].service_name
+  policy       = var.s3_endpoint_policy == "" ? data.aws_iam_policy_document.default.json : var.s3_endpoint_policy
```
bryantbiggs, line 29: I haven't tested myself yet, but I think something like this should work without requiring the default policy data source:
```suggestion
  policy       = var.s3_endpoint_policy == "" ? null : var.s3_endpoint_policy
```

---

one other thing that might work as well (again, haven't tested this here) would be:
1. Change the policy default value to `null` like:
```hcl
variable "s3_endpoint_policy" {
  description = "Custom IAM policy for S3 endpoint"
  type        = string
  default     = null
}
```
2. Then simply do this:
```hcl
  policy       = var.s3_endpoint_policy
```

again, it would need to be tested and validated on either approach but I think it might work and makes it cleaner. I don't know what the AWS provider or AWS API will provide when a null is provided (or no policy is provided)

variables.tf
```diff
@@ -2372,3 +2372,207 @@ variable "create_egress_only_igw" {
...
+  default     = ""
+}
+
+variable "auto_scaling_plans_endpoint_policy" {
+  description = "Custom IAM policy for Auto Scaling Plans VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "workspaces_endpoint_policy" {
```
bryantbiggs, line 2544: looks like workspaces endpoints do not support policies https://docs.aws.amazon.com/vpc/latest/userguide/integrated-services-vpce-list.html

variables.tf
```diff
@@ -2372,3 +2372,207 @@ variable "create_egress_only_igw" {
...
+  description = "Custom IAM policy for Workspaces VPC endpoint"
+  type        = string
+  default     = ""
+}
+
+variable "access_analyzer_endpoint_policy" {
```
bryantbiggs, line 2550: I don't see anything for IAM access analyzer on this list https://docs.aws.amazon.com/vpc/latest/userguide/integrated-services-vpce-list.html

bryantbiggs:
@kostyaplis could you update the [`complete-vpc`](https://github.com/terraform-aws-modules/terraform-aws-vpc/tree/master/examples/complete-vpc) for the changes you are proposing here please? this not only gives others an example on how to configure these modules, but it also is how we test the modules. I would encourage you to update the complete example and try running yourself to make sure everything provisions successfully

kostyaplis:
Hi @bryantbiggs, thanks for your review!

1. Regarding VPC endpoints custom policies availability. 
Unfortunately, AWS documentation is known for some incompleteness and inaccuracies.
So upon developing this PR I had to go VPC -> Endpoints -> Create Endpoint and manually check EACH endpoint for Full Access/Custom policies support. 
I was able to find documentation that mention endpoint policies for some of the endpoints you listed above. While for others I cannot find explicit docs, so you may check yourself (in VPC -> Endpoints -> Create Endpoint) or just trust my word:)

DynamoDB: https://docs.aws.amazon.com/vpc/latest/userguide/vpc-endpoints-ddb.html
ECR DKR: ???
Workspaces: https://docs.aws.amazon.com/workspaces/latest/adminguide/infrastructure-security.html#interface-vpc-endpoint
IAM Access Analyzer: ???

2. Default `aws_iam_policy_document` is there for the reason.
It seems to me that AWS provider treats `null` value for `aws_vpc_endpoint.policy` in some weird way. 
I've implemented and tested changes you suggested.
TF `0.13.5`
AWS Provider `3.16.0`

```hcl
variable "s3_endpoint_policy" {
  description = "Custom IAM policy for S3 VPC endpoint"
  type        = string
  default     = null
}

resource "aws_vpc_endpoint" "s3" {
 ...
  policy = var.s3_endpoint_policy
 ...
}
``` 

Results:
1. Apply module with `enable_s3_endpoint = true` - endpoint correctly created with default `Full Access` policy
2. Declare variable `s3_endpoint_policy` as valid JSON string - endpoint policy correctly updated
3. Set variable `s3_endpoint_policy` back to `null` - no changes generated by tf plan/apply. We'd expect it to revert the policy to its default value
4. Modify `s3_endpoint_policy` in UI or CLI while variable `s3_endpoint_policy = null`. Run terraform plan/apply - no changes generated. We'd expect it to revert the policy to its default value

So that was the reason I put default endpoint policy there. Seems that if declared it has to be valid JSON string and completely skipped if set to `null`.

3. Noted your request on example update. Will try to find some more time this weekend, will ping you once done.

bodgit:
Just came looking through this module repository as I noticed I couldn't set an S3 endpoint policy so I'm keen to see this PR merged so I can use this module instead of our homegrown version. FWIW I've set a DKR endpoint policy.

PR #507, opened 2020-09-22
feat: Added tflint as pre-commit hook

antonbabenko:
Enabling tflint for all directories using the same `.tflint.hcl` config file.


antonbabenko:
[v2.52.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.52.0) has been just released.

PR #517, opened 2020-10-06
revert: Create only required number of NAT gateways

antonbabenko:
Reverts terraform-aws-modules/terraform-aws-vpc#492

The fix was not necessary and not working as expected.


antonbabenko:
[v2.57.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.57.0) has been just released.

See https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/492#issuecomment-704419219 for the explanation.

PR #525, opened 2020-10-16
feat: Added support for Terraform 0.14

antonbabenko:
- fix: Added support for Terraform 0.14
- fix: Removed lifecycle ignore_changes which doesn't work in TF 0.14


antonbabenko:
[v2.58.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.58.0) has been just released.

PR #534, opened 2020-10-25
feat: Lambda VPC Endpoint

aiell0:
## Description
AWS recently announced [PrivateLink support for Lambda](https://aws.amazon.com/blogs/aws/new-use-aws-privatelink-to-access-aws-lambda-over-private-aws-network/). This PR serves to add the functionality to this module. 

## Motivation and Context
So the new endpoint can be used by projects already integrated with this module.

## Breaking Changes
None.

## How Has This Been Tested?
Locally.

### DIFF
--- a/vpc-endpoints.tf
+++ b/vpc-endpoints.tf
@@ -1,5 +1,3 @@
-data "aws_region" "current" {}
-
 ######################
 # VPC Endpoint for S3
 ######################
--- a/vpc-endpoints.tf
+++ b/vpc-endpoints.tf
@@ -190,11 +190,16 @@ resource "aws_vpc_endpoint" "sqs" {
 #########################
 # VPC Endpoint for Lambda
 #########################
+data "aws_vpc_endpoint_service" "lambda" {
+  count = var.create_vpc && var.enable_lambda_endpoint ? 1 : 0
+
+  service = "lambda"
+}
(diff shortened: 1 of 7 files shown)


vpc-endpoints.tf
```diff
@@ -185,6 +187,22 @@ resource "aws_vpc_endpoint" "sqs" {
...
 }
 
+#########################
+# VPC Endpoint for Lambda
+#########################
+resource "aws_vpc_endpoint" "lambda" {
+  count = var.create_vpc && var.enable_lambda_endpoint ? 1 : 0
+
+  vpc_id            = local.vpc_id
+  service_name      = format("com.amazonaws.%s.lambda", data.aws_region.current.name)
```
antonbabenko, line 197: Is there any reason for not using `aws_vpc_endpoint_service` data source to do exactly as most of other VPC endpoints?

If not, please update the code to be as in all other places.

antonbabenko (changes requested):

vpc-endpoints.tf
aiell0: This has been updated to use the data source as indicated.

antonbabenko:
Thanks @aiell0 !

[v2.63.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.63.0) has been just released.

Issue #403, opened 2020-03-09
Private Subnet management is destructive


lgadalla-impinj:
We are creating a set of private subnets, and one of our use cases is the ability to add and remove subnets at will. We notice that the creation of subnets seems to have the "count issue", where if for example, another private subnet is added after the first of three, the second and third are deleted and recreated after the (new) second one is created. Needless to say, this would be disruptive to anyone using the second and third private subnets.

Is there a workaround for this, other than modifying the relevant code to use for_each instead of count?

dpiddockcmp:
Quick answer: no. Terraform does not play well with modifying lists.

Suggestions:
- Only append or delete from the end of the list
- Modify the state file by shifting items in the list before doing plan/apply: `terraform state mv resource[n] resource[n+1]`

In both of these cases you will see knock on issues with other resources not just the subnet. NAT gateways, route tables, routes, etc etc. Terraform also seems to like to trigger full recreation of dependent resources if an item in a counted resource is unknown.

Best option for your usecase would be a forked/new module that uses `for_each`.

achachw:
Hi 
Why don't use a map instead of list to define subnets,to avoid this behaviour
We can for example define a public subnet like this : 
```hcl
public-subnets = {
  subnet1 = {
    cidr = 10.0.11.0/24,
    az    = "eu-west-1b"
  },
  subnet2 = {
    cidr = 10.0.11.0/24,
    az    = "eu-west-1c"
  }
}
```

What do you this about this ?

barryib:
Hi @antonbabenko,

How can we move forward on this ? Can we work on a PR to add map support ? How do we handle breaking changes ? 

Related to https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/178

antonbabenko:
@barryib Good question. Let's discuss it during our own meeting on Friday.

Nickmman:
Hi there, we've run into this same issue and think that using the `for_each` method instead of the `count` method would resolve this issue.

Have talks on this improvement stalled?

bryantbiggs:
there are plans and some preliminary work underway on what a potential v4 *might look like which would look to solve problems such as this - you can track that here if interested https://github.com/clowdhaus/terraform-aws-vpc-v4

Issue #178, opened 2018-10-12
VPC Expansion via secondary_cidr_blocks is incorrectly destructive to original cidr subnets and route tables


tnine:
**Steps to reproduce**

Deploy a simple VPC with the following config.

```yaml
ders = {
    aws = "aws.uw2"
  }

  source = "terraform-aws-modules/vpc/aws"

  name = "auto-vpc-uw2"
  cidr = "10.0.0.0/16"

  azs             = ["us-west-2a", "us-west-2b", "us-west-2c"]
  private_subnets = ["10.0.1.0/24", "10.0.2.0/24", "10.0.3.0/24"]
  public_subnets  = ["10.0.101.0/24", "10.0.102.0/24", "10.0.103.0/24"]

  enable_nat_gateway      = true # needed for SQS communication in mail lambda I think
  enable_dns_hostnames    = true
  map_public_ip_on_launch = true # needed for VPCEs

  private_subnet_tags = {
    type = "private"
  }

  public_subnet_tags = {
    type = "public"
  }
}
```

Plan and apply this configuration to AWS.  Once state is persisted and this exists on AWS, modify to this example

```yaml
module "vpc_uw2" {
  providers = {
    aws = "aws.uw2"
  }

  source = "terraform-aws-modules/vpc/aws"

  name = "auto-vpc-uw2"
  cidr = "10.0.0.0/16"

  secondary_cidr_blocks = [
    "10.1.0.0/16",
    "10.2.0.0/16"
  ]

  azs             = ["us-west-2a", "us-west-2b", "us-west-2c"]
  private_subnets = [
    "10.0.1.0/24", "10.0.2.0/24", "10.0.3.0/24",
    "10.1.1.0/24", "10.1.2.0/24", "10.1.3.0/24",
    "10.2.1.0/24", "10.2.2.0/24", "10.2.3.0/24"
    ]
  public_subnets  = [
    "10.0.101.0/24", "10.0.102.0/24", "10.0.103.0/24",
    "10.1.101.0/24", "10.1.102.0/24", "10.1.103.0/24",
    "10.2.101.0/24", "10.2.102.0/24", "10.2.103.0/24"
    ]

  enable_nat_gateway      = true # needed for SQS communication in mail lambda I think
  enable_dns_hostnames    = true
  map_public_ip_on_launch = true # needed for VPCEs

  private_subnet_tags = {
    type = "private"
  }

  public_subnet_tags = {
    type = "public"
  }
[...]

PR #535, opened 2020-10-26
Fix: Use map instead of list for all subnet type

achachw:
## Description

Defining subnet as list is destructive, when trying to remove one/multiple subnet from that list.

## Motivation and Context

Let's say i want to deploy this configuration :

```hlc
module "vpc-test" {
  ...
  cidr             = "10.40.0.0/16"
  azs              = ["us-east-1a", "us-east-1b"]
  private_subnets  = ["10.40.3.0/24", "10.40.4.0/24","10.40.5.0/24", "10.40.6.0/24"]

```
Once applied, i remove one subnet from that list

```hcl
module "vpc-test" {
  ...
  cidr             = "10.40.0.0/16"
  azs              = ["us-east-1a", "us-east-1b"]
  private_subnets  = ["10.40.3.0/24", "10.40.5.0/24", "10.40.6.0/24"]

```
If i apply this last configuration all private subnets will be destroyed and 3 new ones will be created

#403 
#178 
## Breaking Changes

The idea is to use map instead of list for subnets definitions.
From

```hcl
private_subnets  = ["10.40.3.0/24", "10.40.5.0/24", "10.40.6.0/24"]
```
to
```hcl
private_subnets = {
    "subnet-1" = {
      cidr = "10.40.3.0/24",
      az   = "eu-west-1a"
    },
    "subnet-2" = {
      cidr = "10.40.5.0/24",
      az   = "eu-west-1b"
    },
    "subnet-3" = {
      cidr = "10.40.6.0/24",
      az   = "eu-west-1c"
    }
  }
```
## How Has This Been Tested?

```hcl
module "vpc" {
  source = ".."
  name = "automate-vpc"

  cidr = "20.10.0.0/16"
  enable_nat_gateway = true
  one_nat_gateway_per_az = true

  private_subnets = {
    "subnet-1" = {
      cidr = "20.10.1.0/24",
      az   = "eu-west-1a"
    },
    "subnet-2" = {
      cidr = "20.10.2.0/24",
      az   = "eu-west-1a"
    }
  public_subnets = {
        "subnet-3" = {
          cidr = "20.10.11.0/24",
          az   = "eu-west-1a",
          public_subnet_ipv6_prefixes = 2
        }
     }
  }
```


README.md
```diff
@@ -127,11 +146,76 @@ If both `single_nat_gateway` and `one_nat_gateway_per_az` are set to `true`, the
 By default, the module will determine the number of NAT Gateways to create based on the the `max()` of the private subnet lists (`database_subnets`, `elasticache_subnets`, `private_subnets`, and `redshift_subnets`). The module **does not** take into account the number of `intra_subnets`, since the latter are designed to have no Internet access via NAT Gateway.  For example, if your configuration looks like the following:
 
 ```hcl
-database_subnets    = ["10.0.21.0/24", "10.0.22.0/24"]
-elasticache_subnets = ["10.0.31.0/24", "10.0.32.0/24"]
-private_subnets     = ["10.0.1.0/24", "10.0.2.0/24", "10.0.3.0/24", "10.0.4.0/24", "10.0.5.0/24"]
-redshift_subnets    = ["10.0.41.0/24", "10.0.42.0/24"]
-intra_subnets       = ["10.0.51.0/24", "10.0.52.0/24", "10.0.53.0/24"]
+database__subnets = {
```
aperigault, line 149: Typo

aperigault (commented):
There is many unnecessary new lines in main.tf and README.md

README.md
```diff
@@ -54,10 +54,29 @@ module "vpc" {
...
-  public_subnets  = ["10.0.101.0/24", "10.0.102.0/24", "10.0.103.0/24"]
+  private_subnets = {
+    "subnet-1" = {
+      cidr = "10.0.1.0/24",
+      az   = "eu-west-1a"
+    },
+    "subnet-2" = {
+      cidr = "10.0.2.0/24",
+      az   = "eu-west-1a"
+    }
```
barryib, line 65: @antonbabenko I didn't test it for now, but Is this structure of the subnets variable ok for you ?

antonbabenko: The structure is fine, but I am thinking about an important note on the implementation - aim to have backward compatibility (type `list(string)` should be converted into `map(map(any)`). Have `type = any` on variables.

We can't just ask users to rewrite their code at once even if we release a new major version.

barryib: > The structure is fine, but I am thinking about an important note on the implementation - aim to have backward compatibility (type list(string) should be converted into map(map(any)). Have type = any on variables.

Are you proposing to support for both `list(string)` and `map(map(string))` inputs ? Or to help users to move as easy as possible to the new the input structure (`map(map(string))`) ?

> We can't just ask users to rewrite their code at once even if we release a new major version.

I think we have to document it. We can't do more without adding an extra tool that we should maintain. We can probably provide something for states manipulation, but I'm not sure we want to go through terraform code generation.

antonbabenko: 1. `type = any` (since `type = list(string)|map(map(any))` or similar is not supported). Yes, support both types of values transparently for the users.
2. Agree. We have to document but first, let's try to make it backward compatible. Code generation or even relying on tools like [tfmigrate](https://github.com/minamijoyo/tfmigrate) is not the best option for such modules to my mind.

barryib: > 1. `type = any` (since `type = list(string)|map(map(any))` or similar is not supported). Yes, support both types of values transparently for the users.

I was thinking about using `map(map(string))` and not `map(map(any))`. But using `any` is also ok for me.

> 2. Agree. We have to document but first, let's try to make it backward compatible. Code generation or even relying on tools like [tfmigrate](https://github.com/minamijoyo/tfmigrate) is not the best option for such modules to my mind.

Can you please elaborate your point ? How do you want to handle this backward compatibility without a lot of complexity ? If you have something in mind, please share it.

antonbabenko: 1. `map(map(any))` is better for extending a list of subnets with support for tags (which can be `list(any)` and not `string`). For eg:
```
public_subnet = {
"one" = {
    cidr = "10.0.1.0/24"
    az   = "eu-west-1a"
    tags = ["some", "thing"]
}
...
}
```

2. I have some thoughts, and I will try to put them in writing before our meeting on Friday.

antonbabenko: @barryib 

```terraform
locals {
  one_map = try(tomap(var.subnets), {for k in var.subnets: k => {}})
}

variable "subnets" {
  type = any
  default = ["10.0.0.0/20", "10.0.1.0/20"]
#  default = {
#    "10.0.0.0/20" = { az = "eu-west-1a", tags = { Name: "first" }}
#    "10.0.1.0/20" = { az = "eu-west-1b", tags = { Name: "second" }}
#  }
}
```

`var.subnets` can be list or map, and Terraform will convert that into a map. See around this time - https://youtu.be/xpfGBgLgpWo?t=1520

barryib: Humm.. sounds nice. I'll try that. But this still involve state manipulation right ? It just help users to not modify their code base.

Should we use this for transition only (said for 1 or 2 major release) and we invite them to change their code ?

IMHO, I think that users will be more concerne about state migration and to avoid rebuilding all their infrastructure. The minor code change can be acceptable.

antonbabenko: Yes, manual state manipulation will probably still required, unless we can generate keys to match indexes in a list, like this (to use maps but keep old indexes):

```
one_map = try(tomap(var.subnets), {for k,v in var.subnets: k => { cidr = v }})
```

I think if we use the type conversion, we can keep using it for a long time (more than 1-2 major releases) to support both cases transparently. But we should also see if it is going to look ugly and hard to maintain, we can drop support for older structure (list).

barryib: > Yes, manual state manipulation will probably still required, unless we can generate keys to match indexes in a list, like this (to use maps but keep old indexes):

That was the first thing I tried, but I'm afraid that it won't probably work as expected (but we can still test it). As you mention, map keys are strings and you can't change their type (unless you have an idea).

```shell
# random_pet.node_groups will be destroyed
  - resource "random_pet" "node_groups" {
      - id        = "wondrous-honeybee" -> null
      - keepers   = {
          - "foo" = "10.0.0.0/20"
        } -> null
      - length    = 2 -> null
      - separator = "-" -> null
    }

  # random_pet.node_groups[1] will be destroyed
  - resource "random_pet" "node_groups" {
      - id        = "topical-grizzly" -> null
      - keepers   = {
          - "foo" = "10.0.1.0/20"
        } -> null
      - length    = 2 -> null
      - separator = "-" -> null
    }

  # random_pet.node_groups["0"] will be created
  + resource "random_pet" "node_groups" {
      + id        = (known after apply)
      + keepers   = {
          + "foo" = "10.0.0.0/20"
        }
      + length    = 2
      + separator = "-"
    }

  # random_pet.node_groups["1"] will be created
  + resource "random_pet" "node_groups" {
      + id        = (known after apply)
      + keepers   = {
          + "foo" = "10.0.1.0/20"
        }
      + length    = 2
      + separator = "-"
    }
```

`random_pet.node_groups` => `random_pet.node_groups["0"]`
`random_pet.node_groups[1]` => `random_pet.node_groups["1"]`

Please find bellow my random_pet snippet:

1. I run it first with a `count`
2. I used the magic to convert it to map and used with `for_each`

```hcl
resource "random_pet" "node_groups" {
  for_each = local.one_map
  #count     = length(var.list_subnets)
  separator = "-"
  length    = 2

  keepers = {
    foo = each.value.cidr
    #foo = var.list_subnets[count.index]
  }
}
```

antonbabenko: No surprise here then :) Good that you double-checked it.

barryib:
@antonbabenko how can we move on this ? Is this something you want to add in this module ? Are you planning any general rewrite on the vpc module ? @achachw and I want to help, but need some guidance here.

antonbabenko:
I don't plan to have any rewrites in this module but if you want to work on this PR and experiment with this feature and migrations it would be awesome. At some point, we will have to move forward and introduce migrations natively or using 3rd party tools.

I recommend working on smaller things so that it is easier to read the code.

hansh0801:
track

mustafa89:
Is there any traction on this? we would love to have this feature.

bryantbiggs:
@mustafa89 there is work going on in the background for this but it will take some time. based on prior experience in major changes with modules that are heavily utilized we will have to consider all options

mustafa89:
Alright, looking forward to it, just a small note

```
 "subnet-1" = {
      cidr = "10.40.3.0/24",
      az   = "eu-west-1a"
    },
    "subnet-2" = {
      cidr = "10.40.5.0/24",
      az   = "eu-west-1b"
    },
    "subnet-3" = {
      cidr = "10.40.6.0/24",
      az   = "eu-west-1c"
    }
```
the subnet-1,2,3 feel redundant. I guess a list of maps should also work and simplify it.

```
private_subnets = [
    {
      cidr = "10.40.3.0/24",
      az   = "eu-west-1a"
    },
    {
      cidr = "10.40.5.0/24",
      az   = "eu-west-1b"
    },
    {
      cidr = "10.40.6.0/24",
      az   = "eu-west-1c"
    },
]
```

but yeah, just my 2 cents.

PR #542, opened 2020-12-09
feat: Add outpost support (subnet, NACL, IPv6)

brylex418:
## Description
This adds changes to main / outputs / variables 
To support conditional deployment of outpost subnet(s)

## Motivation and Context
This change is required because creation of subnets on an outpost requires additional arguments to be called (outpost_arn)

## Breaking Changes
None

## How Has This Been Tested?
Yes, tested in account with outpost anchored to it and outpost subnets are created with all other vpc constructs specified.


DrFaust92:
Hey! Thank you for you contribution but im unsure we would want to add support to a feature that would be unmaintainable but the maintainers as im pretty certain we do not have capabilities to test this.

@antonbabenko, thoughts?

antonbabenko:
I agree with @DrFaust92 even if the proposed change looks rather simple we need to be able to run tests periodically.

Does anyone know of any ways to get an outpost emulator locally which we can point [AWS provider endpoint](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/custom-service-endpoints#outposts) onto? Something like [localstack](https://github.com/localstack/localstack) with Outpost support?

brylex418:
At the moment there is not an outposts emulator. 
However we do have an outposts test lab that access could be provided occasionally. As well I have access to outposts.

FireballDWF:
@DrFaust92 @antonbabenko If we setup a CI/CD pipeline under our control which executed the outpost test and emailed you the results, would that work for you?

antonbabenko:
@FireballDWF I don't know how test results can help us in cases related to a service none of the maintainers have access to, especially when users will report bugs or open new feature requests.

Too bad we can't have access to the Outpost sandbox account for a short duration when we need it. This would make our work on this issue much easier.

brylex418:
I can get you in the lab for a couple of days this week. 
@antonbabenko Do you have days that work best for you?

DrFaust92:
IMO it might better to manage outposts resources in a submodule it will encapsulate the logic and reduce the impact on the main repo. @antonbabenko wdyt?

antonbabenko:
There are just 5 resources related to Outpost in this module, and they have their close relatives among existing resources (subnets, network_acl_rule, etc). I don't think it pollutes the module as much as VPC endpoints that can be externalized at some point in the future.

antonbabenko:
@brylex418 Thank you for this PR!

[v2.78.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.78.0) has been just released.

PR #543, opened 2020-12-14
fix: Don't create private route tables when it is not needed

Andor:
## Description

Fix for `aws_route_table.private` creation.

## Motivation and Context

If you are using only `elasticache_subnets` without providing `private_subnets` the "private route tables" will be created anyway. 

## Breaking Changes

With this change, these "private route tables" will be removed.


evairmarinho:
@Andor, take a look at this issue that I opened a little while ago. https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/726. I think something more macro could be done. Because in some cases I don't have NAT Gateway and the module ends up creating N route tables unnecessarily. See if part of the solution I proposed in `local.nat_gateway_count` couldn't make more sense.

PR #147, opened 2018-06-22
Re-adds support for setting tags on the default route table

lorengordon:
Uses a data source to retrieve and preserve the routes
that are currently associated with the route table. This
works around prior problems with supporting this resource,
as described in #95.

Fixes #146

PR #545, opened 2020-12-17
feat: Add support to tag default route table

lays147:
Signed-off-by: Lays Rodrigues <laysrodriguessilva@gmail.com>

## Description
Add support to tag default route table since PR #147 is getting dust.

## Motivation and Context
So far this is the only resource on this module that has not set a Name by default. And I like my resources tagged :3

## Breaking Changes
Not that I know.

## How Has This Been Tested?
Used terraform 0.14
```sh
terraform fmt
terraform validate
terraform plan
terraform apply
```
```hcl
  # module.vpc.aws_default_route_table.default[0] must be replaced
-/+ resource "aws_default_route_table" "default" {
      + default_route_table_id = "rtb-id"
      + id                     = (known after apply)
      + owner_id               = (known after apply)
      + route                  = (known after apply)
      + tags                   = {
          + "Name"                            = "my-rt-name"
          + "environment"                     = "development"
          + "kubernetes.io/cluster/beira-rio" = "shared"
        }
      + vpc_id                 = (known after apply)
    }
```
```
Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```


lays147:
ping @antonbabenko

egarbi:
Have you tried running this over an already created VPC?
Since `default_route_table` is created by default by `aws_vpc` resources, adding this piece of code after, it may cause conflicts trying to create a route that is already created.
If this is true, one would have to use an import to solve the issue, something like:
```bash
terraform import aws_default_route_table.example aws_vpc.this.id
```
This can be tricky for people trying to update the module to a newer version.

daroga0002:
aws_default-route_table is special resource and it is not creating any new route table but is used to pull existing one and manage it using terraform.

As per:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_route_table

```
Each VPC created in AWS comes with a Default Route Table that can be managed, but not destroyed. This is an advanced resource, and has special caveats to be aware of when using it. ```

Issue #435, opened 2020-04-20
CloudWatch LogGroup not destroyed


ahmedfourti:
Hello,

To reproduce this you can create a VPC with thoses inputs 
```  enable_flow_log                           = true  
  create_flow_log_cloudwatch_log_group      = true
  create_flow_log_cloudwatch_iam_role       = true
  flow_log_cloudwatch_log_group_name_prefix = "/my-log-grou-name/" 
```
This will create VPC, enable flow log and create log group as expected.
Then destroy it.
The log group is not destroyed.

However the destroy command succeded.

In reality the log group is destroyed when we issued the destroy command...but as the IAM policy contains `logs:CreateLogGroup` so when a flow log is logged, it will create the log group again.

Do we really need that policy ? In my opinion no as the one who will apply this module has enough permission to create it.

antonbabenko:
Hi,

The IAM policy is for AWS Service "vpc-flow-logs.amazonaws.com" and that service needs to have permissions to create log groups and log streams.

Or I didn't understand what you mean?

ahmedfourti:
Hi,

No this is not needed as the log group is created while deploying the module in the following resource

```resource "aws_cloudwatch_log_group" "flow_log" {
  count = local.create_flow_log_cloudwatch_log_group ? 1 : 0

  name              = "${var.flow_log_cloudwatch_log_group_name_prefix}${local.vpc_id}"
  retention_in_days = var.flow_log_cloudwatch_log_group_retention_in_days
  kms_key_id        = var.flow_log_cloudwatch_log_group_kms_key_id

  tags = merge(var.tags, var.vpc_flow_log_tags)
}```

antonbabenko:
I see, and I think you are right - `"logs:CreateLogGroup"` is not needed in any of two cases (when log group is provided or when it is created by the module).

ahmedfourti:
Yes exactly

PR #550, opened 2021-01-06
Remove CreateLogGroup permission from VPC Flow logs service role

navaati:
## Motivation and Context
This permission is not needed because we create the log group with Terraform so the VPC Flow Logs service doesn’t need to do it. On the other hand having this permission causes a bug where, on terraform destroy the log group will be destroyed, but then if there is still a few messages in a VPC Flow Logs queue the managed service will see that the log group does not exist and create it again using.

You’ll then have the log group lingering after the tf destroy, which can cause trouble if you try to terraform apply again with the same name: the log group will be already existing and your apply will fail. Not having the permission prevents that as the managed service will not be able to recreate the log group after the tf destroy.

## Breaking Changes
Not that I know, except if someone was reusing the created role for another purpose that depended on the permission, which seems unlikely (we don’t expose the role name in the outputs, do we ?).

## How Has This Been Tested?
I’m waiting on CI here https://github.com/trussworks/terraform-aws-vpc-flow-logs/pull/8 which is the exact same change in another module.


bryantbiggs (approved):
looks good - closes #435 - @antonbabenko 👍🏼

antonbabenko:
Thanks @navaati for such a thorough writeup!

[v2.76.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.76.0) has been just released.

PR #560, opened 2021-01-22
feat: vpc endpoint names

cjtiffin:
## Description
This PR adds a `Name` tag to any vpc endpoints that are created by the module.

At the moment vpc endpoints are passed tags from `local.vpce_tags`; this change merges in a `Name` tag (based on `var.name` and the endpoint service's name) following the same pattern used throughout the rest of the module.

## Motivation and Context
This will add more context to any vpc endpoints in the aws console, as well as allowing any testing tools that rely on the `Name` tag (such as [awspec](https://github.com/k1LoW/awspec/blob/master/doc/resource_types.md#vpc_endpoints)) to pick endpoints up for testing.

## Breaking Changes
None known.

## How Has This Been Tested?
Using terraform 0.13.3 I've deployed the module into an aws account with all endpoints enabled; checking that the deployment was successful and that the endpoints were named as I expected.
The endpoints were enabled in two batches, i.e. from `access_analyzer` to `elasticbeanstalk_health`, and then from `elasticloadbalancing` to `workspaces`, as aws has a 50 vpc endpoint limit by default.

### DIFF
--- a/vpc-endpoints.tf
+++ b/vpc-endpoints.tf
@@ -14,8 +14,12 @@ resource "aws_vpc_endpoint" "s3" {
   vpc_id            = local.vpc_id
   service_name      = data.aws_vpc_endpoint_service.s3[0].service_name
   vpc_endpoint_type = var.s3_endpoint_type
-
-  tags = local.vpce_tags
+  tags = merge(
+    {
+      "Name" = format("%s-%s", var.name, data.aws_vpc_endpoint_service.s3[0].service)
+    },
+    local.vpce_tags
+  )
 }
 
 resource "aws_vpc_endpoint_route_table_association" "private_s3" {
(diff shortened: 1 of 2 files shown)


chrismazanec:
perhaps you could reference the data source instead of hard coding the name?
```
  tags = merge(
    {
      "Name" = format("%s-%s", var.name, data.aws_vpc_endpoint_service.s3[0].service)
    },
    local.vpce_tags,
  )
```

bryantbiggs:
thanks for the PR @cjtiffin - I'll take a look today

vpc-endpoints.tf
```diff
@@ -14,8 +14,12 @@ resource "aws_vpc_endpoint" "s3" {
   vpc_id            = local.vpc_id
   service_name      = data.aws_vpc_endpoint_service.s3[0].service_name
   vpc_endpoint_type = var.s3_endpoint_type
-
-  tags = local.vpce_tags
+  tags = merge(
+    {
+      "Name" = format("%s-%s", var.name, data.aws_vpc_endpoint_service.s3[0].service)
+    },
+    local.vpce_tags
```
bryantbiggs, line 21: so this does work but theres one slight issue - if you try to test this using the `examples/complete-vpc` you won't see your changes (all endpoints will still be named `complete`). we have a deeply nested overloading of tags going on currently (`var.tags` -> `var.vpc_endpoint_tags` -> `Name` set here). I think to fix and decouple this change a little better, lets do this:

```hcl
  tags = var.vpce_name_include_service ? merge(
    local.vpce_tags,
    {
      "Name" = format("%s-%s", var.name, data.aws_vpc_endpoint_service.codedeploy[0].service)
    },
  ) : local.vpce_tags
```

And we'll need to add the variable block for the `vpce_name_include_service` variable where the default is `false`. This means that today if users don't want to change their VPC endpoint names (aka - leave them as they have been up till now), then this change is transparent to them. however, if users want to opt into this change and have the endpoint service name appended to the VPC endpoint name, they can flip the switch by setting `vpce_name_include_service = true`

cjtiffin: Yep, good idea. I'll get a new PR put up for this. Thanks

chrismazanec (commented):
hi @bryantbiggs @cjtiffin, shall I make a new direct PR for this or what do you suggest?

comments have been addressed in my branch and PR to @cjtiffin's, https://github.com/cjtiffin/terraform-aws-vpc/pull/2

PR #566, opened 2021-01-29
fix: Fixed wrong count in DMS endpoint

antonbabenko:
Fixed #565

PR #567, opened 2021-01-29
fix: fixing dms endpoint count

andormarkus:
## Description
Sorry Anton,
There was a bug in the dms endpoint code


antonbabenko:
Already fixed in #566 and released [v2.68.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.68.0).

PR #568, opened 2021-01-29
Fix breaking change - Unsuitable value for right operand: bool required.

pnduati:


antonbabenko:
This has been already fixed in [v2.68.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.68.0).

PR #572, opened 2021-02-02
fix: vpc endpoint service s3

andormarkus:
## Description
AWS has some kind of bug on vpc endpoint service S3. S3 should be a "gateway" service however "gateway" and "interface" services shows up. This cause terraform to fail with the following error:

`Error: multiple VPC Endpoint Services matched; use additional constraints to reduce matches to a single VPC Endpoint Service`

Multiple S3 endpoint service (tested region: eu-central-1, eu-west-1)
Screenshot 2021-02-02 at 19 35 45

## Motivation and Context
I made this PR to able to deploy this able to deploy terraform code.

### DIFF
--- a/vpc-endpoints.tf
+++ b/vpc-endpoints.tf
@@ -4,7 +4,9 @@
 data "aws_vpc_endpoint_service" "s3" {
   count = var.create_vpc && var.enable_s3_endpoint ? 1 : 0
 
-  service = "s3"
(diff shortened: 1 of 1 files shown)


discreet:
I am unable to make any VPC changes via Terraform and this module until this is fixed by adding the `service_type`

andormarkus:
Continued in #573

Issue #571, opened 2021-02-02
AWS S3 Endpoints now support Interface type


eriepasquare:
It appears that AWS released (at least in our accounts) a second endpoint type for S3 Endpoints... until now S3 Endpoints only supported Gateway, but now they support an S3 endpoint of type Interface.  Because of this, the vpc module is throwing the following error:

Error: multiple VPC Endpoint Services matched; use additional constraints to reduce matches to a single VPC Endpoint Service

The data source for s3 endpoint needs to include an additional filter to ensure a single record is returned:

https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/a78cee91b4a2c798c0bdabff1d48987bfd89ca25/vpc-endpoints.tf#L4-L8

Should now include:

```terraform
data "aws_vpc_endpoint_service" "s3" {
  count = var.create_vpc && var.enable_s3_endpoint ? 1 : 0

  service = "s3"
  service_type = "Gateway"
}
```

rogersd:
Also related: https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/445

michaelhelmick:
@rogersd thanks for this. Literally, just had a colleague trying to implement something and came across this error. Will have him try it (judging by the 👍🏻) looks like it should work.

vadanx:
Yep. This broke our pipelines. Exact same problem.

123BLiN:
yep, but fix does not work with old aws provider looks like?
```
Error: "service_type": this field cannot be set

  on .terraform/modules/vpc/vpc-endpoints.tf line 4, in data "aws_vpc_endpoint_service" "s3":
   4: data "aws_vpc_endpoint_service" "s3" {
```

123BLiN:
hey @antonbabenko minimal AWS provider version should be increased probably as well, but I've not discovered correct version yet, it is definetly doesn't work with 2.70 (see above)

pib:
They added service_type in 3.10.0

antonbabenko:
Good point, I will update the required version constraint now.

antonbabenko:
#574 => [v2.70.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.70.0) has been just released.

dgonzalez:
So that means it is broken for providers prior to 3.10?

antonbabenko:
Yes, data-source was not available in prior versions.

There are two workarounds we can have in the code:
[...]

Issue #445, opened 2020-05-19
aws data provider Error: multiple VPC Endpoint Services matched


jonathancolby-olx:
versions:

vpc module (this module):  2.33.0
terraform version: v0.12.21
terraform providers: v2.62.0

background info:

we use this TF codebase in 2 different accounts, and 2 different regions.   the error only appears in one of the accounts in eu-west-1.   in this account, we do have 2 custom vpc endpoints (PrivateLink services).   Just a theory, but perhaps the aws_vpc_endpoint_service is incorrectly returning those endpoints, even though they are not S3.  

creating this issue to see if anyone else is having the issue.

```hcl
Error: multiple VPC Endpoint Services matched; use additional constraints to reduce matches to a single VPC Endpoint Service

  on .terraform/modules/vpc/terraform-aws-vpc-2.33.0/vpc-endpoints.tf line 4, in data "aws_vpc_endpoint_service" "s3":
   4: data "aws_vpc_endpoint_service" "s3" {

```

bryantbiggs:
hi @jonathancolby-olx - this error message looks like there are are 2 s3 vpc endpoints and the data filter is unable to return just a single result (its returning multiple matches and the data source does not know how to handle multiple matches, it needs to converge on a single match). currently the filter criteria is just to match on the s3 service https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/26bb020dbc1c240d921964b087adbb1fcfafc758/vpc-endpoints.tf#L7 - if needed, we could look at providing additional filtering functionality but I'm not sure that would solve this exact problem. I could see cases where multiple s3 endpoints are provisioned, but you can only have one s3 endpoint per subnet, so without knowing more its hard to tell what your exact scenario is. do you already have an s3 endpoint deployed?

bryantbiggs:
hi @jonathancolby-olx  - were you able to check if an s3 endpoint already existed in the subnet(s) you are provisioning into?

jonathancolby-olx:
@bryantbiggs -  This issue is resolved.  

The problem was not directly related to the terraform aws provider.  However it  might appear again for other users in the same situation.

Our aws account was enrolled in the beta testing of the new "Interface Endpoint for S3".    The interface endpoint was activated alongside the Gateway Endpoint.

The terraform aws provider of course did not know how to distinguish between the two.  Indeed it is expected that only one aws S3 endpoint is returned.

[...]

PR #573, opened 2021-02-02
fix: Specify an endpoint type for S3 VPC endpoint

haidaraM:
## Description

Correctly filter S3 endpoint based on the type of endpoint.

## Motivation and Context

This is now required because AWS seems to have added a new endpoint Interface for S3. If you don't filter the datasource for S3 endpoint on type, Terraform will fail because it will return multiple VPC endpoints
`Error: multiple VPC Endpoint Services matched; use additional constraints to reduce matches to a single VPC Endpoint Service`

Fix https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/571


andormarkus:
Hi @haidaraM 

I have reviewed your PR I found the following issues:
- Documentation not updated
- Developer can create interface or gateway endpoint not both of them

However I dont see this one as new feature from AWS side. I think it is a bug.
Gateway is free, however interface cost money. I don't see a reason why anyone would use S3 interface instead of gateway.
If you got centralised vpc endpoints I don't see it the reson for it. 

`Gateway endpoints are free to provision, and there is not a strong use case for centralization.` [https://docs.aws.amazon.com/whitepapers/latest/building-scalable-secure-multi-vpc-network-infrastructure/centralized-access-to-vpc-private-endpoints.html](url)

Here is the difference between of the two: [https://gkzz.medium.com/what-is-the-differences-between-vpc-endpoint-gateway-endpoint-ae97bfab97d8](url)

discreet:
I am unable to make any VPC changes via Terraform and this module until this is fixed by adding the service_type

haidaraM:
Hi @andormarkus, Thanks for your feedback:
 - Just updated the documentation
 - It may be a bug and in fact I don't see a reason to use an Interface endpoint here but as it's supported by the API, I added a new variable to support both cases without changing the current behavior people expect

haidaraM:
Just saw this: https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/445#issuecomment-643985043 
Maybe not a bug after all

andormarkus:
Hi @haidaraM 

I assume this is a new interface feature from AWS. I will close my PR #572 

However in this case this PR needs more modification:
- Current integration should only focus on gateway type S3 endpoint. Interface type endpoint should not be associated with routing tables
- It need a separate interface type integration for S3. Here is full commit example for RDS [https://github.com/terraform-aws-modules/terraform-aws-vpc/commit/5446e1c86a6341beef4c58445890c4b9fc96366f](url)
- Lot of documentation in the readme

antonbabenko:
[v2.69.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.69.0) has been just released.

Hopefully, this is going to fix the issue AWS brought to us so unexpectedly.

I also put DynamoDB endpoint type as a configuration in case AWS decides to update it while I sleep :)

Thanks to everyone who worked on the solution for this!

bsakweson:
Here is how I resolved this issue at my end:

- Updated terraform `aws` provider minimum required version to `3.1.0`
- Change my `terraform-aws-vpc` to `v2.70.0`

I saw this error for the first time today and was able to fix it with those two updates.

artburkart:
I couldn't get it to work unless I used AWS Terraform provider `3.10.0`, because the "terraform-aws-vpc" module at version `2.70.0` wouldn't allow me to install an older version.

arik-porat:
Hi guys, how can I get the fix if I'm using terraform version 0.12.20?
This module latest version is only working for terraform version >= 0.12.21

haidaraM:
@arik-porat Update to at least Terraform 0.12.21 and update your AWS provider to at least 3.10

arik-porat:
Going from 0.12.20 to 0.12.21 just require to replace the terrafrom version or there's an upgrade procedure? didn't anything about migrating a minor terraform version.

haidaraM:
From the CHANGELOG, nothing special from 0.12.20 to 0.12.21: https://github.com/hashicorp/terraform/blob/v0.12/CHANGELOG.md#01221-february-19-2020. Just change the version and should work.

If your AWS provider is still on 2.X, you may need to do some changes: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/version-3-upgrade. 

Maybe do the updates and check terraform plan :-)

andormarkus:
Hi @haidaraM,

If you set S3 endpoint type to "Interface" it will cause an error as expected.
`Error: An Interface VPC Endpoint must always have at least one Security Group`

Fixed in #575

haidaraM:
Perfect @andormarkus 

Here is AWS news: https://aws.amazon.com/fr/about-aws/whats-new/2021/02/amazon-s3-now-supports-aws-privatelink/

andormarkus:
Hi @haidaraM 


Typical AWS 😄  They just releasing new features and updating documentation or making announcement later

PR #574, opened 2021-02-02
feat: Upgraded minimum required versions of AWS provider to 3.10

antonbabenko:
Related to #573 since VPC endpoint service_type became available in AWS provider version 3.10.


antonbabenko:
[v2.70.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.70.0) has been just released.

PR #564, opened 2021-01-29
feat: Adding VPC endpoint for DMS

andormarkus:
## Description
Adding missing DMS endpoint

## Motivation and Context
DMS endpoint is missing

PR #575, opened 2021-02-02
feat: s3 interface endpoint

andormarkus:
## Description
Creates the opportunity to create interface style S3 endpoint while it maintains backward API compatibility.
Fixing the error which was introduced in #573 and #564 

## Motivation and Context
In the PR #573 variable "s3_endpoint_type" was introduced. This works with default ("Gateway") value. However if you try to  attach "Interface" interface style endpoint to a route table it will cause error

Used configuration
```  #S3 Endpoint
  enable_s3_endpoint        = true
  enable_public_s3_endpoint = false
  s3_endpoint_type          = "Interface"
```

  
Error message
```
Error: An Interface VPC Endpoint must always have at least one Security Group
```

 <br><br>  
A gateway style endpoint attach itself to a route table
Screenshot 2021-02-02 at 22 20 44
   
 <br><br>
An interface style endpoint attach itself to subnets and needs SG to make it work
Screenshot 2021-02-02 at 22 21 08


bodgit:
Just looking through the existing PR's to see if anyone had added support for this. The S3 interface endpoint doesn't support private DNS though as I just found out.

Issue #585, opened 2021-02-04
issue #582 does not solve the problem with multiple VPC endpoints matched on Terraform 0.12 and aws provider 2.70


dgonzalez:
In the issue it is recommended for people who cannot update the aws provider to version 3.10 to use the release 2.69.0 of terraform-aws-vpc.

I am in such situation (upgrading the provider is not trivial, we are working on it but will take us days if not a couple of weeks) and when I try to plan with the 2.69.0 I get the following error:
```
Error: "service_type": this field cannot be set

  on .terraform/modules/this.vpc/vpc-endpoints.tf line 4, in data "aws_vpc_endpoint_service" "s3":
   4: data "aws_vpc_endpoint_service" "s3" {



Error: "service_type": this field cannot be set

  on .terraform/modules/this.vpc/vpc-endpoints.tf line 45, in data "aws_vpc_endpoint_service" "dynamodb":
  45: data "aws_vpc_endpoint_service" "dynamodb" {
```

This is using the AWS provider 2.70.0. Am I wrong or it is not possible to use this module with a provider below 3.10 after AWS released the S3 PrivateLink interface?

circa10a:
@dgonzalez I have the same problem, here's what I did to get around it until we can upgrade the provider: https://github.com/hashicorp/terraform-provider-aws/issues/17417

tbugfinder:
Hi @circa10a ,
as you have already fixed it, could open open a PR to this repo?
IMHO, fixes for the issue within this repo were applied too quickly as there are means to serve all provider versions.

It would be excellent getting some feedback from maintainers.

bryantbiggs:
hi all, could you try out the changes proposed here https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/597 to see if this resolves the issues you are currently facing please?

tbugfinder:
Hi @bryantbiggs ,
your PR still requires provider v3.10.
We are looking for a fix with provider v2.70.0 compatibility.

https://github.com/bryantbiggs/terraform-aws-vpc/blob/fix/s3-interface-endpoint/versions.tf#L7

bryantbiggs:
ah, you are right @tbugfinder - I forgot that change. I've just pushed that change up after testing/validating

Issue #582, opened 2021-02-03
Solution for "Error: multiple VPC Endpoint Services matched"


antonbabenko:
If you have been using this VPC module and started from 2.2.2021 started seeing the error: `Error: multiple VPC Endpoint Services matched; use additional constraints to reduce matches to a single VPC Endpoint Service`, here are the available solutions depending on the versions you use:

## Terraform 0.12, 0.13, 0.14:

1. Upgrade Terraform AWS provider to be `3.10.0` (see [changelog for the release](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#3100-october-09-2020)) or newer:

```terraform
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = ">= 3.10"
    }
  }
}
```

2. Upgrade version of the module to be `2.70.0` (or newer):

```terraform
module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = "2.70.0"

  # ...
}
```

3. Alternatively, if you can't upgrade Terraform AWS provider as described in the first point, you can use this module version `2.69.0`:

```terraform
module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = "2.69.0"

  # ...
}
```

## Terraform 0.11

You can upgrade your configuration to use `v1.73.0` to get the specified issue fixed on Terraform 0.11 ([commit](https://github.com/terraform-aws-modules/terraform-aws-vpc/commit/691737859e2aaf45e0a84760a87ee9829e0229e3)):

```terraform
module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = "1.73.0"

  # ...
}
```

## Notes

If you want to manage the VPC endpoint for S3 outside of this module (when using Terraform 0.11, for eg), you can set `enable_s3_endpoint = false` and create resources `aws_vpc_endpoint` externally.

## References:

#573

https://github.com/hashicorp/terraform-provider-aws/issues/17417

PR #587, opened 2021-02-06
v2.69.1 proposal, ensure AWS terraform provider v2.70.0 compatibility

circa10a:
## Description

- Remove s3/dynamodb service endpoint data sources due to incompatible `service_type` parameter.
- Replace dynamic service name from data sources with template strings

## Motivation and Context

After this week's release of privatelink for s3, the fixes implemented in v2.69.0 and v2.70.0 don't mitigate errors when using the AWS Terraform provider `v2.X.X`

Addresses:

- #582
- #585 

## Breaking Changes

This is a proposed patch release for `v2.69.` which did not resolve the original issue.

## How Has This Been Tested?

This was tested by creating vpc endpoint services with AWS Terraform Provider `2.70.0` and that no error occurred.

/cc @tbugfinder

### DIFF
--- a/vpc-endpoints.tf
+++ b/vpc-endpoints.tf
@@ -1,11 +1,15 @@
 ######################
 # VPC Endpoint for S3
 ######################
+
 data "aws_vpc_endpoint_service" "s3" {
   count = var.create_vpc && var.enable_s3_endpoint ? 1 : 0
 
-  service_type = var.s3_endpoint_type
-  service      = "s3"
+  service = "s3"
+  filter {
+    name   = "service-type"
+    values = [var.s3_endpoint_type]
+  }
 }
 
 resource "aws_vpc_endpoint" "s3" {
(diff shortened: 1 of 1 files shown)


circa10a:
Any thoughts on this? If none, I'll go ahead and close it.

antonbabenko:
Hi @circa10a ! This looks pretty good, I will take a deeper look at this in a few days and verify how it works.

There is no need to close this one.

bryantbiggs:
hi @circa10a - thanks for the PR! this is a great idea and I have taken this further in https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/597. a couple things to note:
- DynamoDB interface endpoints are currently not supported yet so I have not included that change as you have
- If users were to try to use the S3 interface type, it would fail. I have addressed this as well
- I've updated an example project as well - this is what we use to validate as well as demonstrate changes

circa10a:
Closing in favor of #597

Issue #477, opened 2020-08-01
aws_default_security_group "this" is always dirty when you set manage_default_security_group


apogrebnyak:
Following the example and setting these:
```
  manage_default_security_group  = true
  default_security_group_ingress = [{}]
  default_security_group_egress = [{}]
```
always results in updating `resource "aws_default_security_group" "this"`
```
  ~ resource "aws_default_security_group" "this" {
        arn                    = "arn:reducted"
        description            = "default VPC security group"
      ~ egress                 = [
          + {
              + cidr_blocks      = []
              + description      = ""
              + from_port        = 0
              + ipv6_cidr_blocks = []
              + prefix_list_ids  = []
              + protocol         = "-1"
              + security_groups  = []
              + self             = false
              + to_port          = 0
            },
        ]
        id                     = "sg-reducted"
      ~ ingress                = [
          + {
              + cidr_blocks      = []
              + description      = ""
              + from_port        = 0
              + ipv6_cidr_blocks = []
              + prefix_list_ids  = []
              + protocol         = "-1"
              + security_groups  = []
              + self             = false
              + to_port          = 0
            },
        ]
        name                   = "default"
        owner_id               = "reducted"
        revoke_rules_on_delete = false
        vpc_id                 = "reducted"
    }
```

bluejorts:
any update on this? Mildly annoying to always ignore this change in our TF plans.

pasali:
you can simply use empty array.

    manage_default_security_group  = true
    default_security_group_ingress = []
    default_security_group_egress = []

if you use an array of one empty element, it triggers to execute for_each statement in the resource

```hcl
resource "aws_default_security_group" "this" {
  count = var.create_vpc && var.manage_default_security_group ? 1 : 0

  vpc_id = aws_vpc.this[0].id

  dynamic "ingress" {
    for_each = var.default_security_group_ingress
    content {
      self             = lookup(ingress.value, "self", null)
      cidr_blocks      = compact(split(",", lookup(ingress.value, "cidr_blocks", "")))
      ipv6_cidr_blocks = compact(split(",", lookup(ingress.value, "ipv6_cidr_blocks", "")))
      prefix_list_ids  = compact(split(",", lookup(ingress.value, "prefix_list_ids", "")))
      security_groups  = compact(split(",", lookup(ingress.value, "security_groups", "")))
      description      = lookup(ingress.value, "description", null)
[...]

PR #591, opened 2021-02-11
fix: aws_default_security_group was always dirty when manage_default_security_group was set

schollii:
The example uses `[{}]` for the `default_security_group_ingress` and `default_security_group_egress` which causes the issue described in #477. The solution described there has worked for me and several others so this pull request fixes the example so others don't waste time having to figure this out.


bryantbiggs:
thanks @schollii - I will take a look today

bryantbiggs:
this is just a change to the `examples/complete-vpc` and does work as intended, safe to merge @antonbabenko 👍🏼 - thanks @schollii

PR #592, opened 2021-02-11
chore: Adds database_subnet_group_name as an output variable

DeLoWaN:
## Description
This PR adds an output variable which points to the name of the database subnet group name

## Motivation and Context
We need to access this variable since the [terraform-aws-modules/rds/aws](https://registry.terraform.io/modules/terraform-aws-modules/rds/aws/latest) modules needs the name of the database subnet group when it already exists (and not its ID). Although I would prefer that this module use the ID as input, it's much more simpler to add an output to the VPC module.

At the moment, we can't just assume the database subnet group name with the `name` input parameter since a `lower()` function is applied, and may change it the future.

## Breaking Changes
No breaking changes


outputs.tf
```diff
@@ -133,6 +133,11 @@ output "database_subnet_group" {
   value       = concat(aws_db_subnet_group.database.*.id, [""])[0]
 }
 
+output "database_subnet_group_name" {
+  description = "Name of database subnet group"
+  value       = aws_db_subnet_group.database.*.name[0]
```
DrFaust92, line 138: do we need the `[0]` at the end?

DrFaust92 (changes requested):

outputs.tf
DeLoWaN: We don't want a list here, but a string. And ss the database subnet group name [is the same](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/main.tf#L411) whatever the number of subnets is defined, it is safe to take the first occurrence on the list.

bryantbiggs, line 138: lets keep it consistent:
```suggestion
  value       = concat(aws_db_subnet_group.database.*.name, [""])[0]
```

bryantbiggs (approved):
looks good, @antonbabenko 👍🏼

antonbabenko:
@DeLoWaN Thanks!

[v2.73.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.73.0) has been just released.

PR #584, opened 2021-02-03
docs: update the 'conditional creation' section of the README for TF 0.13+

ae-ou:
…hat modules support the 'count' argument (as of Terraform 0.13).

## Description
Updated the 'Conditional Creation' section of the documentation to reflect the fact that Terraform 0.13+ supports the count argument in module blocks (https://www.terraform.io/docs/language/meta-arguments/count.html).

## Motivation and Context
The documentation was outdated and didn't fully apply to newer versions of Terraform.

## Breaking Changes
No breaking changes.

## How Has This Been Tested?
I read it, and my IDE has a spellchecker?

PR #597, opened 2021-02-22
fix: use filter for getting service type for S3 endpoint and update to allow s3 to use interface endpoint types

### LINKED ISSUE #578: Broken dependency since #574 - incorrect use of provider source for 0.12 versions of Terraform
Hi,

#574 

This commit: https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/574/files#diff-7a675c670d75c5377fab042389a1c31e147037138af03515f2d455e357aaef60

The above commit is fundamentally flawed. A provider source (in this case for AWS) cannot be supplied for versions of Terraform < 0.13 yet the constraint for the required version of Terraform still permits 0.12 versions of Terraform.

Anyone who depends on this resource who is still running a 0.12 version of Terraform will now find themselves with a broken implementation.

I suspect this was tested using a 0.13 version of Terraform. Try again with a 0.12 version to reproduce.

Thanks.

### LINKED ISSUE #585: issue #582 does not solve the problem with multiple VPC endpoints matched on Terraform 0.12 and aws provider 2.70

bryantbiggs:
## Description
- use filter for getting service type for S3 endpoint and update to allow s3 to use interface endpoint types
- reduce min supported version of AWS provider back to 2.70.0 with these changes

## Motivation and Context
- there was a change made to avoid conflicts between the newly introduced S3 `Interface` endpoint and instead keep the module at the default setting of `Gateway` for backwards comparability. However, this caused two issues:
1. Users need to upgrade their version of the AWS provider to have the attribute type `s3_endpoint_type` available to them
2. If users were to switch to an `Interface` type for S3, it would fail

This PR attempts to solve both of those problems and the `examples/complete-vpc` has been updated to validate and demonstrate this change.

Ref: #582 
Closes #575
Closes #578
Closes #585 
Closes #587 

## Breaking Changes
- No

## How Has This Been Tested?
1. Checkout `master` locally at time of writing
2. `cd examples/complete-vpc && terraform init && terraform apply`
3. Checkout this PR locally
4. `terraform apply` -> S3 endpoint type successfully changes to type `Interface` following the norms established
5. I also re-applied `master` to set back to "current" and again checked out my PR but setting the example back to what it is currently on `master` and no changes which means that backwards compatibility should be preserved

I also tested by setting the `examples/simple-vpc` AWS version to `v2.70.0` and validating that it works at that version as well still


examples/complete-vpc/main.tf
```diff
@@ -118,8 +121,8 @@ module "vpc" {
 
   # Default security group - ingress/egress rules cleared to deny all
   manage_default_security_group  = true
-  default_security_group_ingress = [{}]
-  default_security_group_egress  = [{}]
+  default_security_group_ingress = []
```
bryantbiggs, line 124: changing these two lines is related to #584

bryantbiggs:
Note - I had to remove the changes from #573 for the DynamoDB endpoint since `Interface` is not supported by DynamoDB yet. There are two routes we could take here:
1. Leave the change in for DynamoDB but update it like I have here for S3 knowing that if a user tries to provision an endpoint type of `Interface` for DynamoDB that it will not work
2. `[Currently this is the route I've taken]` Roll back the changes for DynamoDB and plan to update similar to what I've done here for S3 when it is made available by AWS

vpc-endpoints.tf
```diff
@@ -45,18 +54,15 @@ resource "aws_vpc_endpoint_route_table_association" "public_s3" {
 data "aws_vpc_endpoint_service" "dynamodb" {
   count = var.create_vpc && var.enable_dynamodb_endpoint ? 1 : 0
 
-  service_type = var.dynamodb_endpoint_type
-  service      = "dynamodb"
+  service = "dynamodb"
```
antonbabenko, line 57: I think we can add `service-type` filter as you do for S3 VPC endpoint to be forward-compatible with the upcoming DynamoDB in case it happens anytime (users won't have to do anything to get it to work and it won't break their setup).

bryantbiggs: sounds good - just pushed up the update for this

antonbabenko:
This was a big one!

[v2.74.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v2.74.0) has been just released.

bodgit:
So what happens if you want to use both types of S3 endpoint? (Which is a perfectly valid scenario)

Also, you've exposed a setting for enabling private DNS on the S3 interface endpoint, which isn't supported by AWS.

bryantbiggs:
@bodgit - I don't believe you can use both endpoint types (though I could be wrong), though I'd be curious why you think you need both types supported. And yes, we have exposed attributes that AWS does not support so that things are "future-proof" (i.e. - DynamoDB interface type is not supported yet but we've set it up here the same as the S3 changes in anticipation, and yes the private DNS on S3 interface endpoints is currently not supported). 

This is just a by-product of trying to make a declarative tool/framework handle multiple use cases (flexibility) - we try to provide sane defaults and expose as much functionality as possible but its up to users to select the appropriate levers and configurations that they need.

bodgit:
Yes, you can use both, they're complimentary. See https://docs.aws.amazon.com/AmazonS3/latest/userguide/privatelink-interface-endpoints.html for examples.

bryantbiggs:
ah ok, thanks for pointing that out - PRs are always welcome! here we were trying to fix some issues that came about with the rollout by AWS and still maintain the current status quo, but I can see where additional changes could be made to support both types

bodgit:
I was watching #575 which seemed a better way of treating both endpoint types as separate settings, given you don't add the interface-type endpoint to routing tables and the gateway endpoint doesn't have a security group, etc.

bryantbiggs:
as I said, we can add that functionality in but we had a situation where users were forced to upgrade to v3.10.0 of the AWS provider or risk running into issues because of the rollout by AWS for this change. the change here allows them to fix their current setup at the lower v2.70.0 version, and if we go forward with whats proposed in #575 then we have to upgrade the min version of AWS provider to v3.10.0

PR #605, opened 2021-03-02
shouldnt this be 20.10.0.2

tomahawk-pilot:


antonbabenko:
The values provided in examples are sometimes wrong or irrelevant in real cases. Users should consult official AWS documentation if in doubt.

PR #609, opened 2021-03-03
feat: Optionally create separate nat gateway subnets

justincmoy:
## Description
If a list of NAT Gateway subnets is specified, generate the NAT Gateway(s) in those subnets.

## Motivation and Context
This enables architecture pieces, like a Network Firewall, to live in the public subnet: https://docs.aws.amazon.com/network-firewall/latest/developerguide/arch-igw-ngw.html

## Breaking Changes
This does not break backwards compatibility as the resources are gated behind a `length(var.nat_gateway_subnets) > 0` check, no existing variables are modified, and no existing outputs are modified.

## How Has This Been Tested?
It was tested with:
```
enable_nat_gateway     = true
one_nat_gateway_per_az = true
create_igw             = false
```

With and without a value for `nat_gateway_subnets`. And resources to add a network firewall to the public subnet.

It was also tested with:
```
enable_nat_gateway     = true
one_nat_gateway_per_az = true
create_igw             = true
```

With and without a value for `nat_gateway_subnets`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -617,6 +617,12 @@ No Modules.
 | monitoring\_endpoint\_subnet\_ids | The ID of one or more subnets in which to create a network interface for CloudWatch Monitoring endpoint. Only a single subnet within an AZ is supported. If omitted, private subnets will be used. | `list(string)` | `[]` | no |
 | name | Name to be used on all the resources as identifier | `string` | `""` | no |
 | nat\_eip\_tags | Additional tags for the NAT EIP | `map(string)` | `{}` | no |
+| nat\_gateway\_route\_table\_tags | Additional tags for the nat gateway route tables | `map(string)` | `{}` | no |
+| nat\_gateway\_subnet\_assign\_ipv6\_address\_on\_creation | Assign IPv6 address on nat\_gateway subnet, must be disabled to change IPv6 CIDRs. This is the IPv6 equivalent of map\_public\_ip\_on\_launch | `bool` | `null` | no |
+| nat\_gateway\_subnet\_ipv6\_prefixes | Assigns IPv6 nat\_gateway subnet id based on the Amazon provided /56 prefix base 10 integer (0-256). Must be of equal length to the corresponding IPv4 subnet list | `list(string)` | `[]` | no |
+| nat\_gateway\_subnet\_suffix | Suffix to append to public subnets name | `string` | `"nat-gateway"` | no |
+| nat\_gateway\_subnet\_tags | Additional tags for the nat gateway subnets | `map(string)` | `{}` | no |
+| nat\_gateway\_subnets | A list of nat gateway subnets inside the VPC | `list(string)` | `[]` | no |
 | nat\_gateway\_tags | Additional tags for the NAT gateways | `map(string)` | `{}` | no |
 | one\_nat\_gateway\_per\_az | Should be true if you want only one NAT Gateway per availability zone. Requires `var.azs` to be set, and the number of `public_subnets` created to be greater than or equal to the number of availability zones specified in `var.azs`. | `bool` | `false` | no |
 | private\_acl\_tags | Additional tags for the private subnets network ACL | `map(string)` | `{}` | no |
@@ -799,6 +805,7 @@ No Modules.
 | intra\_subnets\_cidr\_blocks | List of cidr\_blocks of intra subnets |
 | intra\_subnets\_ipv6\_cidr\_blocks | List of IPv6 cidr\_blocks of intra subnets in an IPv6 enabled VPC |
 | name | The name of the VPC specified as argument to this module |
+| nat\_gateway\_route\_table\_ids | List of IDs of nat gateway route tables |
 | nat\_ids | List of allocation ID of Elastic IPs created for AWS NAT Gateway |
 | nat\_public\_ips | List of public Elastic IPs created for AWS NAT Gateway |
 | natgw\_ids | List of NAT Gateway IDs |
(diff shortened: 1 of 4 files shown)


dojci:
Hi @justincmoy could you please update your PR ? This functionality looks interesting, so i can do a review for you.

DrFaust92:
possibly consider https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/660#issuecomment-895561033

antonbabenko:
I am rather skeptical to have just another type of subnet introduced into this module unless it is critical.

Why can't you use a public subnet where NAT gateway will be created and network firewall subnet will be using it? 

Referring to the AWS documentation link:
1. `NAT gateway subnet` = public subnet
2. `Customer subnet` = private subnet`
3. `Firewall subnet` = the new one, which should be created as a part of Network Firewall infrastructure (I guess) once VPC is created.

Am I missing something?

dojci:
> I am rather skeptical to have just another type of subnet introduced into this module unless it is critical.
> 
> Why can't you use a public subnet where NAT gateway will be created and network firewall subnet will be using it?
> 
> Referring to the AWS documentation link:
> 
>     1. `NAT gateway subnet` = public subnet
> 
>     2. `Customer subnet` = private subnet`
> 
>     3. `Firewall subnet` = the new one, which should be created as a part of Network Firewall infrastructure (I guess) once VPC is created.
> 
> 
> Am I missing something?

@antonbabenko Sometimes you don't want to mix up services like NAT into one subnet together with other workloads. There are multiple use-cases when it doesn't fit:
1. Firewall actually "lives" between NAT GW (subnets) and internet gateway. In a setup where NAT GW = public subnet, you are forcing traffic from public subnets to goes through the firewall event we are only interested in NAT outgoing traffic.
2. You want to use dedicated ACLs, route tables, routes, ...
3. You are checking/forcing compliance rules (acls, routing rules, ...)
4. There are much more complicated network setups where simply patterns cannot be used

antonbabenko:
I agree with what you say and specifically the statement ("There are much more complicated network setups where simply patterns cannot be used") makes me eager to close this PR and continue asking users to create extra resources outside of this module if they have advanced use-cases.

I just can't imagine how we can handle the accidental complexity in this module over time if we continue adding support for niche and/or advanced use-cases.

/cc @DrFaust92 @bryantbiggs

dojci:
> I agree with what you say and specifically the statement ("There are much more complicated network setups where simply patterns cannot be used") makes me eager to close this PR and continue asking users to create extra resources outside of this module if they have advanced use-cases.
> 
> I just can't imagine how we can handle the accidental complexity in this module over time if we continue adding support for niche and/or advanced use-cases.
> 
> /cc @DrFaust92 @bryantbiggs

I don't think that this PR makes code really complex, and there is an added value as I wrote in the previous post. We are using something similar in a separate module but this is a temporary solution (we rely on PR merge). This temp solution in separate module is a hack and it's really pain to do this way there are many problems you can hit. NAT GW as a resource is deeply integrated with VPC and as @justincmoy said there are couple of connections:
* IGW must be created before NAT GWs, currently module doesn't have the option to create IGW without public subnets
* There is no way to use NAT GW from the VPC module with target subnets that are created outside of the VPC module 
* Multiple routing rules inside module depends on NAT gw (default route table, private route table, db route table), NAT GW in a separate module creates the requirement to manage routing rules outside of VPC module
* Single vs Multiple NAT GWs

I understand your concerns about complexity over time, but this PR introduces many possibilities for alternative network setups and not only for the one specific case without big overhead.

antonbabenko:
@justincmoy will you be able to update this PR with only these changes:
1. Add a variable to customize the subnet used for NAT gateways?
2. Add a note in README describing how users can use this module to provision part of the network infrastructure required for the Network Firewall. Or maybe add an end-to-end example in `examples` directory if you can.

dojci:
@DrFaust92 @antonbabenko Can you please explain what you mean by "override the list of subnets on the NAT GW itself" ? I feel like I'm missing something. Are you proposing to create subnets outside of the VPC module and use these subnets as the input parameter to the NAT GW resource ? If so, how / when they will be created (thinking about subnet vpc_id dependency) ?

DrFaust92:
@dojci i meant this line https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/609/files#diff-dc46acf24afd63ef8c556b77c126ccc6e578bc87e3aa09a931f33d9bf2532fbbR1055

regarding deps. terraform manages deps on resource level and not module levels
so if you create a subnet (outside of the module) using vpc id from module output and then use subent ids from the external subnets it should work fine.

dojci:
> @dojci i meant this line https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/609/files#diff-dc46acf24afd63ef8c556b77c126ccc6e578bc87e3aa09a931f33d9bf2532fbbR1055
> 
> regarding deps. terraform manages deps on resource level and not module levels
> so if you create a subnet (outside of the module) using vpc id from module output and then use subent ids from the external subnets it should work fine.

yes if you are in the same tf run. Perhaps, custom NAT subnets input will be sufficient and easy to impement / maintain without shortcomings.

jrstarke:
> 1. `NAT gateway subnet` = public subnet
> 2. `Customer subnet` = private subnet`
> 3. `Firewall subnet` = the new one, which should be created as a part of Network Firewall infrastructure (I guess) once VPC is created.

This is the architecture that has been followed in the Network Firewall PR (https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/672)

PR #629, opened 2021-04-12
feat: add subnet and route table outputs suitable for use with for_each

### LINKED ISSUE #628: Outputs for subnets and route tables are not well suited for use with for_each
No

The various subnet/route_table outputs are purely lists of IDs, which are not knowable until after the apply has run.  This results in them being poorly suited for use in a resource that makes use of the for_each functionality (the key should be something that is computable before the apply, the values can be unknown until after apply).  This results in errors such as the following if you attempt to do a for_each using, for example, one of the subnet arn outputs.

```
Error: Invalid for_each argument

  on modules/workload_vpc/subnet_share.tf line 112, in resource "aws_ram_resource_association" "database_subnets":
 112:   for_each = toset(module.vpc.database_subnet_arns)

The "for_each" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the for_each depends on.
```

The module should output a proper map for the elements where the key is something that is computable before the apply (I would recommend either the CIDR block or the value of the Name tag for naming the subnet keys and the value of the Name tag for the route table

Currently I am having to do a two-stage apply where first I apply just this VPC module and then after that I can pull data sources based on the existing subnet ids and subnet arns resources to then iterate over in a resource using a for_each loop

## Additional context
I am more than happy to write the PR for this change as I have already tested it locally and it's pretty simple to accomplish.  Stand by for an incoming PR.

cryptk:
## Description
This change adds outputs for aws_subnet and aws_route_table resources which are stored in a map suitable for use in a for_each.

## Motivation and Context
Resolves #628 .  This module is for the most part not well suited for use along with for_each.  This PR is a potential first-step in resolving those issues.  This will prevent errors such as those documented in #628 (although I am sure there are similar cases for other resources).

## Breaking Changes
This should not break any backwards compatibility as it only adds some new outputs to the module.

## How Has This Been Tested?
I created an issue-628-subnets-for-each example which creates various subnets and then iterates over them to create aws_ram_resource_share resources for each subnet type.


cryptk:
Working to clean it up a little and fix the test failure

examples/issue-628-subnets-for-each/main.tf
```diff
@@ -0,0 +1,57 @@
...
+
+  single_nat_gateway = true
+  enable_nat_gateway = true
+
+  enable_s3_endpoint       = true
+  enable_dynamodb_endpoint = true
+}
+
+resource "aws_ram_resource_share" "private_subnets" {
+  for_each                  = module.vpc.private_subnets_map
```
antonbabenko, line 29: You can also replace this for_each with:
```
for_each = { for k,v in module.vpc.public_subnets : k => v }
```

It will work as excepted (no need to do anything with outputs).

antonbabenko: I know this does not solve the issue of having type lists that can be solved if we use maps, but this PR is not moving the needle towards us using maps in both places (in variables and in outputs).

antonbabenko (commented):
Thanks for the PR but this is won't help us in the process of improving from the `list` to `map` in both variables and outputs.

examples/issue-628-subnets-for-each/main.tf
cryptk: @antonbabenko you cannot replace that for_each with what you mentioned.  The issue that I opened in #628 has the reasoning why.  You cannot do a for_each across a map whos keys are not known at plan time.  `module.vpc.public_subnets` is a list of subnet IDs, those IDs are not known at plan time unless the subnets have already been created, which makes them not suitable for use as the key in a map which is referenced in a for_each.

The workaround for this is to apply in two steps, first to apply the `module.vpc` to create the VPC and the subnets and then to apply the rest of the code (the subnets IDs would be known at plan time on the second run as they would already exist).

This PR resolves that issue by outputting the data from the module in the form of a map whos keys CAN be computed at plan time (because they are based off information that is not based on a return from the AWS API, but rather information that is supplied to terraform directly).  This removes the requirement to do a two-step apply where you first apply the VPC module to ensure that the information being used in the map keys is located in the state.  This is the problem that your proposed solution does not resolve.

cryptk: While I do agree that this does not fully resolve the issue of the module receiving information as a list and being based off counts (lists) for resource creation, it does resolve the issue of outputs not being maps.  This PR should be viewed as a "first step" and the format of these outputs could be maintained for compatibility pretty easily as the internals of the module are reworked to build the resources based on maps/for_each allowing for consumers of the module more time to migrate their code to be based off of these map based outputs and avoid a compatibility crunch when the module is eventually reworked.

cryptk: @antonbabenko did my statements above clarify what this PR resolves and why it is different than the issue that you are speaking about?  This functionality would greatly increase the usability of this module in some more advanced scenarios.

antonbabenko: I am not sure because I didn't have time to go very deep since I wrote my comment, run it, and it worked 3 days ago. It will require some time to get through this PR.

jplock:
I just do something like:

```
resource "aws_ram_resource_association" "public_subnets" {
  for_each           = toset(module.vpc.public_subnet_arns)
  resource_arn       = each.key
  resource_share_arn = aws_ram_resource_share.organization.arn
}
```

cryptk:
> I do something like:
> 
> ```terraform
> resource "aws_ram_resource_association" "public_subnets" {
>   for_each           = toset(module.vpc.public_subnet_arns)
>   resource_arn       = each.key
>   resource_share_arn = aws_ram_resource_share.organization.arn
> }
> ```

That will work fine if the subnets already exist (if the subnets don't exist yet, terraform cannot know the ARNs).  The only way to guarantee that the subnets exist is to do a two step apply where you first apply the subnets and then afterwards you apply your example RAM resource.  If the subnets already exist, you could skip the initial apply, but you wouldn't be able to do a single apply "from scratch".

jpalmerpivotal:
Hi, I have been checking in on this PR periodically. I want to clarify what the status with respect to this PR/the underlying issue is.

From issue #628 it appears that issue #403 was considered a possible umbrella issue for this type of change, but from the talk on this thread I am unclear as to whether this PR is being considered as a stopgap.

It looks like #403 is not actively being worked on.

So questions:
1. Is #403 something that might happen in the near future?
2. If #403 is not being worked on is this PR being considered as a stopgap?

PR #647, opened 2021-05-07
fix: it is impossible to set private_dns_enabled as true for S3

kei500:
## Description
- update README of sub-module `vpc-endpoints`.

## Motivation and Context
- `private_dns_enabled` cannot be specified for S3 endpoint.

## Breaking Changes
- no

## How Has This Been Tested?
- README change only.

### DIFF
--- a/modules/vpc-endpoints/README.md
+++ b/modules/vpc-endpoints/README.md
@@ -17,7 +17,6 @@ module "endpoints" {
     s3 = {
       # interface endpoint
       service             = "s3"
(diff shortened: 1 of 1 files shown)


bryantbiggs:
hi @kei500 - private DNS can be enabled when using the [S3 interface endpoint](https://docs.aws.amazon.com/AmazonS3/latest/userguide/privatelink-interface-endpoints.html), but not with the gateway endpoint

PR #648, opened 2021-05-18
fix: README.md for VPC Endpoints and use proper service names

evilr00t:
Based on the output of:

```sh
 aws ec2 describe-vpc-endpoint-services --region us-east-1
```

Services names are using different naming convention, they use URL like naming with region included.

Hope that helps!

### DIFF
--- a/modules/vpc-endpoints/README.md
+++ b/modules/vpc-endpoints/README.md
@@ -7,6 +7,8 @@ Terraform sub-module which creates VPC endpoint resources on AWS.
 See [`examples`](../../examples) directory for working examples to reference:
 
 ```hcl
+data "aws_region" "current" {}
+
 module "endpoints" {
   source = "terraform-aws-modules/vpc/aws//modules/vpc-endpoints"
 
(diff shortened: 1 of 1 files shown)


bryantbiggs:
this is incorrect - we pass the name as a var to the [`aws_vpc_endpoint_service`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/vpc_endpoint_service) data source which will determine what the fully qualified name of the endpoint should be. this also allows for custom endpoints as well as AWS provided service endpoints

evilr00t:
@bryantbiggs understood, I think I have some edge-case then:

```module "vpc_endpoints" {
  source = "terraform-aws-modules/vpc/aws//modules/vpc-endpoints"

  vpc_id             = module._vpc.vpc_id
  security_group_ids = [module.all_local.this_security_group_id]

  endpoints = {
    s3 = {
      # interface endpoint
      service             = "s3"
      service_type        = "Gateway"
      private_dns_enabled = true
      route_table_ids     = module._vpc.private_route_table_ids
      tags                = { Name = "s3-vpc-endpoint" }
    },
    dynamodb = {
      # gateway endpoint
      service         = "dynamodb"
      # service_type    = "Gateway"
      # service_name    = "com.amazonaws.${data.aws_region.current.name}.dynamodb"
      route_table_ids = module._vpc.private_route_table_ids
      tags            = { Name = "dynamodb-vpc-endpoint" }
    }
  }

  tags = {
    Application = local.app_name
    Environment = local.env
  }
}
```

Gives me:

```│ Error: no matching VPC Endpoint Service found
│
│   with module.us-east-1.module.vpc_endpoints.data.aws_vpc_endpoint_service.this["dynamodb"],
│   on .terraform/modules/us-east-1.vpc_endpoints/modules/vpc-endpoints/main.tf line 9, in data "aws_vpc_endpoint_service" "this":
│    9: data "aws_vpc_endpoint_service" "this" {
│
```

and

`service_type    = "Gateway"`

should be added for dynamodb example anyway.

evilr00t:
Ok - looks like just service_type is the thing here - will update commit.

Issue #612, opened 2021-03-07
Implementing `aws_network_acl_rule` with count causes rule deletions if order of rules change


2stacks:
## Description

Inserting a new rule or changing rule order causes acls that haven't changed to be deleted and re-added.  Should use for_each instead of count.

bryantbiggs:
hi all, this should be addressed in the next version which is under development here https://github.com/clowdhaus/terraform-aws-vpc-v4/tree/main/modules/subnets

2stacks:
Awesome thanks for the update.

PR #649, opened 2021-05-19
fix: use for_each instead count for aws_network_acl_rule resources

jrhrmsll:
## Description

- Use `for_each` instead of `count` for `aws_network_acl_rule` resources.

## Motivation and Context

This change try to solve issues that occur when new  Network ACL Rules are inserted among existing ones, causing some of them to be missing after the apply.

- Closes: [#612](https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/612)

## Breaking Changes

- No

## How Has This Been Tested?
Tested by using examples/network-acls.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -692,37 +692,37 @@ resource "aws_network_acl" "public" {
 }
 
 resource "aws_network_acl_rule" "public_inbound" {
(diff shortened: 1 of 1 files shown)


vtomasr5 (approved):

DrFaust92:
This module was created prior to for_each being available thus the count. changing now from count to for_each is a breaking change that will cause a recreate. while more future proof this will need to be considered fro next major version of the module

jeffadams (approved):

PR #651, opened 2021-06-07
chore: Private DNS cannot be used with S3 endpoint

lazzurs:
Private DNS is not a valid setting for the S3 interface endpoint.

## Description
Updating documentation.

## Motivation and Context
The documentation for this module is invalid.

## Breaking Changes
None

## How Has This Been Tested?
Have tried to run the code as provided in the documentation and it will not work.


antonbabenko (approved):
Thanks for the finding and correction!

PR #653, opened 2021-06-09
feat: Add support for EKS secondary cidr range

j3ffrw:
## Description
Adds capability to a another subnet with NAT GW that can be used by EKS worker nodes for its secondary IP addresses
(AWS)[https://aws.amazon.com/about-aws/whats-new/2018/10/amazon-eks-now-supports-additional-vpc-cidr-blocks/]  

## Motivation and Context
Provisioning EKS worker nodes on a /24 subnet will consume a lot of address space to address this AWS added support for secondary CIDR range for EKS. 

## Breaking Changes
No breaking change, this just add a new subnet resource named for EKS 

## How Has This Been Tested?
- I have tested and validated these changes using one or more of the provided `examples/eks-secondary-cidr` projects
- Using the example above I created a simple vpc first without the secondary cidr then added the secondary cidr range afterwards.

[eks_secondary_cidr_test.log](https://github.com/terraform-aws-modules/terraform-aws-vpc/files/6625004/eks_secondary_cidr_test.log)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 <a name="unreleased"></a>
 ## [Unreleased]
 
+- feat: Add EKS secondary CIDR support ()[https://aws.amazon.com/about-aws/whats-new/2018/10/amazon-eks-now-supports-additional-vpc-cidr-blocks/]
+
 
 
 <a name="v3.1.0"></a>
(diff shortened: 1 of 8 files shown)


antonbabenko:
I am not sure I understand the premise for this PR. You can already specify `secondary_cidr_blocks` to have a bigger VPC, and you can set CIDR larger than `/24` in subnets.

I don't think we need to have just another type of subnets and related resources specifically for EKS.

Am I missing something?

j3ffrw:
hi, yes that is an option and we've used them before but I would like to have an option to also separate them especially when I need to run a lot of pods but can't use a subnet bigger than /24 on the primary/private subnet. Would you prefer to have this subnet   made or named as a custom subnet that can be reused for other purposes aside from EKS?

antonbabenko:
There is no need to make custom subnets (and related resources) in this module because it will be very complex to handle all the relations in the code. If you need to make extra resources, please handle them outside of the module.

Closing this issue.

PR #662, opened 2021-07-13
Ap vpc endpoint s3

darkside720:
## Motivation and Context

Error: multiple VPC Endpoint Services matched; use additional constraints to reduce matches to a single VPC Endpoint Service

## Breaking Changes

No

## How Has This Been Tested?
- [ ] I have tested and validated these changes using one or more of the provided `examples/*` projects local project

Updated VPC S3 endpoint and tested on local env which completed successfully

Plan: 0 to add, 0 to change, 0 to destroy.

No


bryantbiggs:
I'm sorry I don't think this is valid as its using an older version of the module

PR #664, opened 2021-07-15
fix: Fixed mistake in separate private route tables example

hieultan:
## Description

Fix copy/paste mistake of the outputs in VPC with separate private route tables example
## Motivation and Context

This still working but maybe misleading 
## Breaking Changes

I don't think any of these are breaking changes, just fix copy/paste mistake. No existing properties are being changed.
## How Has This Been Tested?


DrFaust92 (approved):
LGTM

PR #666, opened 2021-07-22
feat: Added transit_gateway subnet support

dbarrachina-ntt:
## Description
Allow transit gateway subnets to be created independently. 

## Motivation and Context
[AWS transit gateway best practices](https://docs.aws.amazon.com/vpc/latest/tgw/tgw-best-design-practices.html) suggest that transit gateway attachments should have their own dedicated /28 subnet, so you can apply different policies on them.

## Breaking Changes
No

## How Has This Been Tested?
Tested with `complete-vpc` example.
Additionally we use this module and we are creating transit gateway subnets using this code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,14 +83,16 @@ If both `single_nat_gateway` and `one_nat_gateway_per_az` are set to `true`, the
 
 ### One NAT Gateway per subnet (default)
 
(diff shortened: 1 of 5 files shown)


dbarrachina-ntt:
Detected missing some missing outputs, will create a new one shortly

PR #672, opened 2021-07-30
feat: Added resources for aws network firewall

### LINKED ISSUE #665: Subnets for AWS Network Firewall with NAT and IGW use-cases
Yes, AWS Network Firewall

It doesn't appear possible to use the existing module to setup AWS network firewall in the way AWS described where the firewall sits in dedicated subnets (with nothing else in them) and traffic from NAT goes to firewall before going to IGW. Example: https://docs.aws.amazon.com/network-firewall/latest/developerguide/arch-igw-ngw.html

Perhaps a dedicated subnet type for AWS network firewall? 

## Additional context
I'm seeing more use cases in which AWS network firewall is being requested and due to its unique subnet and routing requirements it may require a new subnet type.

ryanoolala:
## Description

Fixes #665 

Adding support fire AWS Network firewall. Creating the new subnet type called "firewall_subnets" and modifying the required route tables for traffic to go through the firewall endpoint

## Motivation and Context

AWS Network Firewall was released almost a year ago, it is a feature that my organisation will like to make use off, and seeing how it requires changes in route tables that are created and managed in this module, e.g the public subnet to internet gateway route, i've decided to modify the vpc module so that the changes are easily made, without having to diverge my terraform state for those changes.

Network ACLs are not able to handle domain blocks as well a large number of IP blocks due to the limited number of NACL rules you can create.

## Breaking Changes

None, i've tested it with the complete-vpc example before and after my changes. 

## How Has This Been Tested?

The resources for firewall will only be created if the following variables are provided

- `firewall_subnets` (creates the subnets and route table for the new subnet to IGW)
- `enable_firewall` (creates the firewall service, modifies the public subnet to route through the firewall endpoint, and creates the IGW ingress route table to firewall endpoints)

I've added an example on setting this up, as well as a diagram drawn with app.diagram.net, with the drawing metadata embedded for easy modifications in the future if required.


examples/network-firewall/main.tf
```diff
@@ -0,0 +1,125 @@
...
+  firewall_policy_arn = aws_networkfirewall_firewall_policy.example.arn
+  vpc_id              = module.vpc.vpc_id
+
+  dynamic "subnet_mapping" {
+    for_each = module.vpc.firewall_subnets
+    content {
+      subnet_id = subnet_mapping.value
+    }
+  }
+}
```
jrstarke, line 60: It seems like this two step process exists because the firewall is outside of the module. What if you moved the firewall into the module? You could still keep the policy outside the module.

Full disclosure, I happen to be working on something very similar, so perhaps we should collaborate :)

ryanoolala: oh hey thanks for the suggestion, i think that's possible, also seeing that the firewall is placed in the `VPC` services on the console, seems fine for it to be created in a `VPC` module as well. 

I didn't think about doing that because i am not sure if this module should be creating a network firewall, but this cyclic dependency does make it seem like it belong together. Should be a small change, i can add the resource soon

ryanoolala: actually,  i remember why it is not included together now, if i include aws_networkfirewall_firewall, i will also have to include aws_networkfirewall_resource_policy, not sure what is a good base policy that can apply for everyone

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/networkfirewall_firewall#firewall_policy_arn

rasmus:
Nice, we really need this as well 😄

examples/network-firewall/main.tf
jrstarke: You don't actually need to include the network firewall policy, I pass the arn in as a variable: https://github.com/terraform-aws-modules/terraform-aws-vpc/compare/master...jrstarke:network-firewall?expand=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R180

I'm going to see if I can make a pull request against your branch

main.tf
```diff
@@ -196,23 +197,27 @@ resource "aws_default_route_table" "default" {
 ################################################################################
 
 resource "aws_route_table" "public" {
-  count = var.create_vpc && length(var.public_subnets) > 0 ? 1 : 0
+  count = var.create_vpc && length(var.public_subnets) > 0 && length(var.firewall_sync_states) == 0 ? 1 : length(var.firewall_sync_states) > 0 ? local.nat_gateway_count : 0
```
jrstarke, line 200: Why does the number of public routes depend on the number of NAT Gateways?

main.tf
```diff
@@ -1043,6 +1132,59 @@ resource "aws_network_acl_rule" "elasticache_outbound" {
...
+    var.tags,
+    var.firewall_acl_tags,
+  )
+
+  lifecycle {
+    ignore_changes = [
+      tags["Owner"],
+      tags["Type"],
+    ]
+  }
```
jrstarke, line 1157: Changes to network ACL doesn't require replacement (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-network-acl.html). Why are we ignoring these?

main.tf
```diff
@@ -602,6 +662,35 @@ resource "aws_subnet" "intra" {
   )
 }
 
+################################################################################
+# Firewall subnet
+################################################################################
+resource "aws_subnet" "firewall" {
+  count = var.create_vpc && length(var.firewall_subnets) > 0 && (false == var.one_nat_gateway_per_az || length(var.firewall_subnets) >= length(var.azs)) ? length(var.firewall_subnets) : 0
```
jrstarke, line 669: I'm a little confused where the `var.one_nat_gateway_per_az` fits into this. From [AWS' post on Deployment Models for Network firewall](https://aws.amazon.com/blogs/networking-and-content-delivery/deployment-models-for-aws-network-firewall/), if you're using a NAT then they still expect that the NATs will be in the protected subnets, which will connect to the Network Firewall

jrstarke:
I've put up a branch that builds on yours, adding firewall logs, and return path through the network firewall. I was trying to do it as a pull request, but it wasn't working. You can check it out here: https://github.com/terraform-aws-modules/terraform-aws-vpc/compare/master...jrstarke:feature/aws-network-firewall?expand=1

main.tf
ryanoolala: ah my bad, will remove these

main.tf
ryanoolala: hmm this works well when `one_nat_gateway_per_az=true` because each nat gateway will be in its own protected subnet, and for each zone in the variable `az` so if there user only provides `['1a','1b']` and the region actually has a `1c` region, there will only be 2 NAT gateways, and implies there should be 2 firewall subnets supplied, and 2 public(1a and 1b) route tables, to map to each firewall endpoint. 

Basically i'm reusing the calculation for nat_gateway_count as i think it covers the scenarios best, but i'm not sure what kind of vpc designs people do with this module. 

Could there be a case where there are >1 nat gateway in a availability zone?

jrstarke: Ah interesting. From the [AWS Network Firewall Deployment Models](https://aws.amazon.com/blogs/networking-and-content-delivery/deployment-models-for-aws-network-firewall/) model 2 that supports NAT Gateways, it appears that the Private subnets need to route to the NAT, and the Public/Protected subnets need to point to the Network Firewall's VPC Endpoint. [Check out their diagram here](https://d2908q01vomqb2.cloudfront.net/5b384ce32d8cdef02bc3a139d4cac0a22bb029e8/2020/11/18/anfw-public-igw-deployment-high-res1-1-1.png). So in my commit, I changed this to be [one route table per public subnet](https://github.com/terraform-aws-modules/terraform-aws-vpc/commit/c3eca3e0434177d4743c2ee91d364b2193c24989#diff-dc46acf24afd63ef8c556b77c126ccc6e578bc87e3aa09a931f33d9bf2532fbbR219), so that we could route it to the associated Network Firewall VPC endpoint.

main.tf
ryanoolala: i think i'll follow what you did to make it simpler, i was overthinking and following the count from `aws_subnet.public`

I'll try doing a reverse PR from your branch to mine for this.

ryanoolala:
> I've put up a branch that builds on yours, adding firewall logs, and return path through the network firewall. I was trying to do it as a pull request, but it wasn't working. You can check it out here: https://github.com/terraform-aws-modules/terraform-aws-vpc/compare/master...jrstarke:feature/aws-network-firewall?expand=1

Created the [PR](https://github.com/ryanoolala/terraform-aws-vpc/pull/1)

jrstarke:
This looks good to me. Thanks for all your work @ryanoolala

README.md
```diff
@@ -171,6 +171,16 @@ Sometimes it is handy to have public access to Redshift clusters (for example if
 
 It is possible to integrate this VPC module with [terraform-aws-transit-gateway module](https://github.com/terraform-aws-modules/terraform-aws-transit-gateway) which handles the creation of TGW resources and VPC attachments. See [complete example there](https://github.com/terraform-aws-modules/terraform-aws-transit-gateway/tree/master/examples/complete).
 
+## Network Firewall Configuration
+If you need to use the AWS Network Firewall to provide ingress or egress restrictions that can't be accommodated easily
+by security groups, you can enable it using the following arguments:
+```hcl
+enable_networkfirewall     = true
+networkfirewall_subnets    = ["10.0.11.0/24", "10.0.12.0/24", "10.0.13.0/24"]
+networkfirewall_policy_arn = "arn:aws:network-firewall:<region>:<account_id>:firewall-policy/<policy_name>"
```
fishpen0, line 180: These vars in this doc don't match the vars in the `variables.tf`

```
variable "enable_firewall" {
  description = "Should be true if you want to provision a network firewall in front of your public networks"
  type        = bool
  default     = false
}
...
variable "firewall_subnets" {
  description = "A list of firewall subnets inside the VPC"
  type        = list(string)
  default     = []
}
...
variable "firewall_policy_arn" {
  description = "The network firewall policy arn to associate with the network firewall. Needed if you are setting enable_firewall to true"
  type        = string
  default     = null
}
```

fishpen0 (changes requested):

main.tf
```diff
@@ -1205,9 +1369,24 @@ resource "aws_route_table_association" "public" {
   count = var.create_vpc && length(var.public_subnets) > 0 ? length(var.public_subnets) : 0
 
   subnet_id      = element(aws_subnet.public.*.id, count.index)
-  route_table_id = aws_route_table.public[0].id
+  route_table_id = aws_route_table.public[var.enable_firewall ? count.index : 0].id
+}
+
+resource "aws_route_table_association" "public_internet_gateway" {
+  count = var.create_vpc && var.create_igw && length(var.firewall_subnets) > 0 ? 1 : 0
```
fishpen0, line 1376: IMO, Some loon (me) should be able to create the firewall subnets and leave `enable_firewall` set to false.  This needs to be changed to make that possible as `aws_route_table.internet_gateway` is not created unless `enable_firewall` is true.
```
  count = var.create_vpc && var.create_igw && var.enable_firewall && length(var.firewall_subnets) > 0 ? 1 : 0
```

fishpen0 (changes requested):

README.md
```diff
@@ -171,6 +171,16 @@ Sometimes it is handy to have public access to Redshift clusters (for example if
 
 It is possible to integrate this VPC module with [terraform-aws-transit-gateway module](https://github.com/terraform-aws-modules/terraform-aws-transit-gateway) which handles the creation of TGW resources and VPC attachments. See [complete example there](https://github.com/terraform-aws-modules/terraform-aws-transit-gateway/tree/master/examples/complete).
 
+## Network Firewall Configuration
+If you need to use the AWS Network Firewall to provide ingress or egress restrictions that can't be accommodated easily
+by security groups, you can enable it using the following arguments:
+```hcl
+enable_networkfirewall     = true
+networkfirewall_subnets    = ["10.0.11.0/24", "10.0.12.0/24", "10.0.13.0/24"]
```
fishpen0, line 179: Consider changing these to /28s as this is what amazon documentation suggests.  /28s are also already used in the example files elsewhere in the diff.

network-firewall-logs.tf
```diff
@@ -0,0 +1,39 @@
+locals {
+  enable_firewall_logs = var.create_vpc && var.enable_firewall && var.enable_firewall_logs
+}
+
+########################
+# Network Firewall Logs
+########################
+resource "aws_networkfirewall_logging_configuration" "this" {
+  count = local.enable_firewall_logs && length(var.firewall_log_types) > 0 ? 1 : 0
+
```
fishpen0, line 10: I get a transient error running this:

```
Error: 2 errors occurred:
        * error adding Logging Configuration to NetworkFirewall Firewall (arn:aws:network-firewall:us-east-1:123456789123:firewall/shared2-wallets-api-eks-vpc-firewall): ResourceNotFoundException: LogDestination: /aws/network-firewall-log/vpc-******************-flow does not exist
        * error adding Logging Configuration to NetworkFirewall Firewall (arn:aws:network-firewall:us-east-1:123456789123:firewall/shared2-wallets-api-eks-vpc-firewall): InvalidRequestException: Given logging configuration attempts to create/modify multiple log destination configs



  on vpc/network-firewall-logs.tf line 8, in resource "aws_networkfirewall_logging_configuration" "this":
   8: resource "aws_networkfirewall_logging_configuration" "this" {


Releasing state lock. This may take a few moments...
```

It looks like the `aws_networkfirewall_logging_configuration.this` resource needs a `depends_on` for `aws_cloudwatch_log_group.firewall_log`

fishpen0 (changes requested):

variables.tf
```diff
@@ -208,6 +226,12 @@ variable "intra_subnets" {
   default     = []
 }
 
+variable "firewall_subnets" {
+  description = "A list of firewall subnets inside the VPC"
+  type        = list(string)
+  default     = []
+}
```
fishpen0, line 233:
```
Error: error creating NetworkFirewall Firewall (blah-blah-firewall): InvalidRequestException: availability zone subnet(s) cannot be mapped more than one per availability zone in a firewall, parameter: [subnet-00000000000000001, subnet-00000000000000002]

  on ../../lib/vpc/main.tf line 130, in resource "aws_networkfirewall_firewall" "this":
 130: resource "aws_networkfirewall_firewall" "this" {
 ```

There is a hard requirement for `length(var.firewall_subnets) <= length(var.azs)` imposed by aws.  The rest of this module does not behave this way outside of when the firewall is enabled. (AKA you can have 6 public subnets and 3 azs and other configurations)

This should be documented.

File: variables.tf [hunk 10]
```diff
@@ -1032,6 +1142,36 @@ variable "enable_flow_log" {
...
+  description = "Specifies the name prefix of Network Firewall Log Group for Network Firewall logs."
+  type        = string
+  default     = "/aws/network-firewall-log/"
+}
+
+variable "firewall_log_cloudwatch_log_group_retention_in_days" {
+  description = "Specifies the number of days you want to retain log events in the specified log group for Network Firewall logs."
+  type        = number
+  default     = null
+}
```
fishpen0, line 1161:
```
  ~ resource "aws_cloudwatch_log_group" "firewall_log" {
        arn               = "arn:aws:logs:us-east-1:123456789123:log-group:/aws/network-firewall-log/vpc-022a76fc738dbd151-flow"
        id                = "/aws/network-firewall-log/vpc-00000000000000001-flow"
        name              = "/aws/network-firewall-log/vpc-00000000000000001-flow"
      ~ retention_in_days = 120 -> 0
        tags              = {
            "kubernetes.io/cluster/shared2-wallets-api-eks" = "shared"
        }
        tags_all          = {
            "kubernetes.io/cluster/shared2-wallets-api-eks" = "shared"
        }
    }
```
A default of `null` causes this var to not be marked as required in the docs automator, but will cause TF to constantly re-create this resource.  aws defaults to 120 days, update that to this value or something else.

ryanoolala:
Thanks for reviewing, i'll look through and update as required!

README.md
ryanoolala: thanks i've updated the readme for this

main.tf
ryanoolala: Thanks, i've made the changes to include it into the condition

README.md
ryanoolala: updated thanks

network-firewall-logs.tf
ryanoolala: Have included this, thank you

variables.tf
ryanoolala: Added note in variable description, as well as main README to highlight this point

variables.tf
ryanoolala: Updated default to 120

andreybudzar:
Hi, any updated on this PR merge?

ryanoolala:
@antonbabenko have you had a chance to look at this PR yet?

antonbabenko:
@ryanoolala Not yet. Maybe I will look at it on the 19th of November during [my live stream](http://bit.ly/terraform-youtube).

ryanoolala:
> @ryanoolala Not yet. Maybe I will look at it on the 19th of November during [my live stream](http://bit.ly/terraform-youtube).

oh cool didn't know you have a youtube

jseiser:
So this will not be getting merged, correct?

haarchri:
can we see this merged ? what are the open points ?

andreybudzar:
> can we see this merged ? what are the open points ?

Hi...we'd also like to see this merged. Are there any updates?

ryanoolala:
Hmm so as i watched the preliminary review on youtube, i understand a few questions that have popped up as
1. Why is firewall resource created in this VPC module and not separately
2. Including other deployment methods instead of just supporting one type as mentioned in the PR description

I'll try to answer to help clarify these points @antonbabenko 

I originally separated the firewall resource from this module in my initial commits, this requires the VPC module to be terraformed twice, as the first time will require the `firewall_subnets` to be created in this module, followed by creating the firewall resource in its separate module using the created `firewall_subnets`, and coming back to input the firewall outputs as input in this module in order to create the proper route tables
This was bought up in the community review to include the firewall resource creation, to eliminate this problem.

I did not include the various deployment methods highlighted by the AWS blog post as i'm not familiar with transit gateway etc and do not have an existing infrastructure to test this out on, i was hoping maybe the community can build on top of this to support the other models, if it is possible to do so with this VPC module. Also since it is already a rather large change, I don't wish to add on more complexity than this already is creating. 

Having said that, i understand why this is taking some time for review, although i do hope to merge this in soon so i may move my projects' module reference away from my personal repository to the community one again :)
Hope you can find some time to review it through, let me know if you need any more clarifications or any other ways you think of collaborating that will help accelerate this as well!

jseiser:
The entire subnet/route issue is where we got to as well, when attempting this.  Multiple VPC's created using this module, then the need for the Firewall came, and it became more and more difficult  to create the firewall resources and then manage the already created routes.

main.tf
```diff
@@ -226,6 +250,18 @@ resource "aws_route" "public_internet_gateway" {
   }
 }
 
+resource "aws_route" "public_firewall" {
+  count = var.create_vpc && var.enable_firewall ? length(var.public_subnets) : 0
+
+  route_table_id         = element(aws_route_table.public.*.id, count.index)
+  destination_cidr_block = "0.0.0.0/0"
+  vpc_endpoint_id        = element([for ss in tolist(aws_networkfirewall_firewall.this[0].firewall_status[0].sync_states) : ss.attachment[0].endpoint_id if ss.attachment[0].subnet_id == aws_subnet.firewall[count.index].id], 0)
```
wiegandf, line 258: Because of this, it requires to specify at least as many firewall_subnets as public_subnets. 

Should we default to first firewall_subnet in case aws_subnet.firewall[count.index] doesn't exist?

ryanoolala: Hello, are u able to share a terraform example of your configuration has issues?

I think i understand your first sentence, but not sure how defaulting to first subnet helps.

wiegandf: Hi @ryanoolala, sure - this is a simple example that fails:
```
  cidr = "10.10.0.0/18"

  azs             = ["us-east-1a", "us-east-1b", "us-east-1c"]
  public_subnets  = ["10.10.4.0/24", "10.10.5.0/24"]
  private_subnets = ["10.10.8.0/21", "10.10.16.0/21", "10.10.24.0/21"]

  enable_firewall      = true
  firewall_subnets     = ["10.10.1.0/28"]
  firewall_policy_arn  = dependency.firewall-policy.outputs.arn

  single_nat_gateway = true
  enable_nat_gateway = true
```

will lead to 
```
╷
│ Error: Invalid index
│
│   on main.tf line 258, in resource "aws_route" "public_firewall":
│  258:   vpc_endpoint_id        = element([for ss in tolist(aws_networkfirewall_firewall.this[0].firewall_status[0].sync_states) : ss.attachment[0].endpoint_id if ss.attachment[0].subnet_id == aws_subnet.firewall[count.index].id], 0)
│     ├────────────────
│     │ aws_subnet.firewall is tuple with 1 element
│     │ count.index is 1
│
│ The given key does not identify an element in this collection value: the
│ given index is greater than or equal to the length of the collection.
```

ryanoolala: Hello, in this case you have only 1 firewall endpoint in `us-east-1a`, public in 1a,1b and private in 1a,1b,1c

Is there a reason why you have your setup in such a way, where firewall is not highly available, while the private is the most highly available one? because if 1a goes down with your firewall, you will not have any path to internet anymore, and typically if you are already using multi az setups, you would want to maintain that across scenarios.

wiegandf: Indeed, that's more an artificial example. But I think it could be a valid use case to use more public subnets than firewall subnets. E.g. this

```
  azs             = ["us-east-1a", "us-east-1b", "us-east-1c"]
  public_subnets  = ["10.10.4.0/24", "10.10.5.0/24", "10.10.6.0/24", "10.10.7.0/24", "10.10.8.0/24"]
  private_subnets = ["10.10.8.0/21", "10.10.16.0/21", "10.10.24.0/21"]
  firewall_subnets     = ["10.10.1.0/28", "10.10.2.0/28", "10.10.3.0/28"]
```

So this could be a setup for certain users and it wouldn't make sense to have more than three firewall subnets here.

ryanoolala: hmm yes i think it possible to have more public subnets, I'm also thinking if its possible to have more firewall subnets than the number of AZs in general, which i don't feel it is good to have even though it does help in cost savings, but defeats the purpose of having multi availability zones.

My current thoughts is that the firewall should support scenarios where
`length(var.public_subnets) > length(var.firewall_subnets) && length(var.azs) == length(var.firewall_subnets)`

I'll try to find some time to work on this

deploymentking: Just to let you know @ryanoolala there's a team here in the UK really appreciating the work you're doing on this feature 🥇

ryanoolala: Hello @wiegandf, I've made the changes to support additional public subnets, let me know if it works for you?

@deploymentking Thank you for the appreciation! I'm glad to contribute to the community.

AkselAllas: This functionality would be a godsend for us! :) Currently manually reworking a bunch of code to make it work :thinking:

bryantbiggs:
FYI - this is being worked on here https://github.com/clowdhaus/terraform-aws-vpc-v4

Which includes new resources such as:
- https://github.com/clowdhaus/terraform-aws-vpc-v4/tree/main/modules/network-firewall
- https://github.com/clowdhaus/terraform-aws-vpc-v4/tree/main/modules/dns-firewall
- https://github.com/clowdhaus/terraform-aws-vpc-v4/tree/main/modules/subnets which allow for any number of subnets and subnet configurations

ryanoolala:
> FYI - this is being worked on here https://github.com/clowdhaus/terraform-aws-vpc-v4
> 
> Which includes new resources such as:
> 
>     * https://github.com/clowdhaus/terraform-aws-vpc-v4/tree/main/modules/network-firewall
> 
>     * https://github.com/clowdhaus/terraform-aws-vpc-v4/tree/main/modules/dns-firewall
> 
>     * https://github.com/clowdhaus/terraform-aws-vpc-v4/tree/main/modules/subnets which allow for any number of subnets and subnet configurations

Ah i see, looks like this isn't needed anymore then. Hope to see the module get released soon.

sjpalf:
@ryanoolala please don't give up on this PR yet, this functionality is definitely needed in this module. We need to implement AWS Network Firewall and it will be very difficult for us to move away from using this module. The related PR is for a repo with 0 forks compared to the 2.9k that this repo has!
@antonbabenko are you going to be able to re-review this PR at any time?

antonbabenko:
@sjpalf The larger work is in progress in https://github.com/clowdhaus/terraform-aws-vpc-v4 repository by @bryantbiggs and we will merge it here as soon as it is implemented&tested there. There is a lot of work to be done before we can just "add network firewall resources". Sorry that it takes so long time.

sjpalf:
> @sjpalf The larger work is in progress in https://github.com/clowdhaus/terraform-aws-vpc-v4 repository by @bryantbiggs and we will merge it here as soon as it is implemented&tested there. There is a lot of work to be done before we can just "add network firewall resources". Sorry that it takes so long time.

my apologies, I totally misunderstood what the clowdhaus repo was for. Thanks for your time and effort on this

PR #681, opened 2021-08-17
feat: Added device_name to customer gateway object.

byjg:
## Description
Minimal change adding the `device_name` property into the `aws_customer_gateway` resource.

## Motivation and Context
I already had a Customer Gateway created, and when I imported the resource state into this module, terraform was trying to recreate the resource. Without this change, I  cannot use the module with pre-existent AWS customer gateway resources with the "device_name" set. 

## Breaking Changes
There are no break changes. 

## How Has This Been Tested?
I have two Gateway resources. One with the `device_name` property set, and the other don't set. 

```
    customer_gateways = {
      "cgw1"      = {
        bgp_asn     = 65000
        ip_address  = "1.2.3.4"
        device_name = "USG"
      },
      "cgw2" = {
        bgp_asn    = 65000
        ip_address = "5.6.7.8"
      },
    }
```

I tested both importing the resources and created the resources from scratch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,13 +189,13 @@ It is possible to integrate this VPC module with [terraform-aws-transit-gateway 
 | Name | Version |
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.12.31 |
-| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 3.15 |
+| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 3.28 |
 
 ## Providers
 
 | Name | Version |
 |------|---------|
-| <a name="provider_aws"></a> [aws](#provider\_aws) | >= 3.15 |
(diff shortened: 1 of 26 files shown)


bryantbiggs:
thanks for the PR @byjg - this looks good! could you update the example similar to what you have shown above please https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/2ee908db287a834e263559ec4b35ebac625a7dba/examples/complete-vpc/main.tf#L47-L56

byjg:
@bryantbiggs example added. Thanks

antonbabenko (changes requested):
Please update `versions.tf` file here and in example (probably) since `device_name` was introduced in v3.28.0 of Terraform AWS provider.

byjg:
@antonbabenko version updated.

antonbabenko:
Thanks, @byjg !

[v3.6.0](https://github.com/terraform-aws-modules/terraform-aws-vpc/releases/v3.6.0) has been just released.

PR #688, opened 2021-08-31
feat: Add support for naming and tagging subnet groups

juho9000:
## Description
Add support for elasticache_subnet_group_name, elasticache_subnet_group_tags and redshift_subnet_group_name similar to what database subnet group already has

## Motivation and Context
It is expected that naming and tagging of subnet groups work similarily in each case

## Breaking Changes
Should not break anything

## How Has This Been Tested?


main.tf
```diff
@@ -569,9 +569,17 @@ resource "aws_subnet" "elasticache" {
 resource "aws_elasticache_subnet_group" "elasticache" {
   count = var.create_vpc && length(var.elasticache_subnets) > 0 && var.create_elasticache_subnet_group ? 1 : 0
 
-  name        = var.name
+  name        = lower(coalesce(var.elasticache_subnet_group_name, var.name))
   description = "ElastiCache subnet group for ${var.name}"
   subnet_ids  = aws_subnet.elasticache.*.id
+
+  tags = merge(
```
antonbabenko, line 576: Make sure to update `versions.tf` accordingly. `tags` were added in v3.38.0 of Terraform AWS provider ([link](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#3380-april-30-2021))

antonbabenko (changes requested):

juho9000:
Force pushed some changes related to naming, I had names passed to lower() originally (as they are in database_subnet_group), but then quickly realized that it might cause breaking changes for updates. Sadly database_subnet_group name is passed through lower() so one might expect it's how naming works in all subnet groups.

antonbabenko:
`lower()` is used because AWS API expects it to be in lowercase all the time. I think it is good now.

Thanks for the PR.

juho9000:
#372 can probably be closed aswell, seems to be a zombie PR.

Issue #466, opened 2020-06-30
private_subnet_suffix and public_subnet_suffix do not work


juris:
I'm trying to use this module to create a VPC. Everything is working except suffixes for private/public subnets. Public and private subnets just get tags I've defined in locals section.
```
module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = "v2.44.0"

  name = var.name

  cidr            = var.cidr
  azs             = var.azs
  private_subnets = var.private_subnets
  public_subnets  = var.public_subnets

  private_subnet_suffix = "int"
  public_subnet_suffix  = "ext"

  enable_dns_hostnames = true
  enable_dns_support   = true

  enable_nat_gateway = true
  single_nat_gateway = true

  tags = local.tags
}
```

```
➜ terraform version
Terraform v0.12.28
+ provider.aws v2.68.0
```

gregmoy:
db and elasticache do not work either.

gregmoy:
https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/475

DrFaust92:
Hey @gregmoy, does the `local.tags` map contains a tag with the key `Name` by any chance?

gregmoy:
@DrFaust92 Yeah we were using the tags from https://github.com/cloudposse/terraform-null-label and it appears to include Name as one of the keys. 

It still doesn't make sense to let the tags override Name though. I can't think of a case where someone would want all of their assets named the same exact thing...

DrFaust92:
I know we have some overrides in our implementation, i'll check and get back to you. you might be right and i'm missing something here.

gugaiz:
If you are using the examples provided here, probably #495 will solve your issue

nahueloyha:
Hi, I'm having this exact same issue, just created this [PR](https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/712) that seems to solve it.

PR #712, opened 2021-11-23
fix: name tag merge order 

nahueloyha:
## Description
When passing a tags map with a "Name" key, the custom name generated by this module (appending the "private/public/db" suffix to the subnets name, for example) is ignored.

This happens specially when using in conjunction with the [terraform-null-label](https://github.com/cloudposse/terraform-null-label) module, which is already exporting the tags map with "the value of the Name key is set to the module id output", when passing directly the tags from one module to the other (`tags = module.terraform_null_label.tags`).

## Motivation and Context
This PR fixes issue #466, by changing the order in which the tags are merged. 

As per Terraform documentation over the [merge](https://www.terraform.io/docs/language/functions/format.html) function, "if more than one given map or object defines the same key or attribute, then the one that is later in the argument sequence takes precedence."

Therefore, when the tags map used as an input of this module already contains a "Name" key, the merge will keep that one instead of using the generated one with format.

## Breaking Changes
No, it's backward compatible as it does not break nor recreate anything; it will just update the tags (including the "Name" one used as the resource human-friendly ID).

## How Has This Been Tested?
This was applied on an existing config, with just the expected change on affected resources managed by this module, like this one on a public subnet:

```
  # module.base_vpc.aws_subnet.public[0] will be updated in-place
  ~ resource "aws_subnet" "public" {
        id                              = "subnet-0fdca660bc0b7d881"
      ~ tags                            = {
          ~ "Name"      = "my-subnet-name" -> "my-subnet-name-public-us-east-1a"
            # (6 unchanged elements hidden)
        }
      ~ tags_all                        = {
          ~ "Name"      = "my-subnet-name" -> "my-subnet-name-public-us-east-1a"
            # (6 unchanged elements hidden)
        }
        # (9 unchanged attributes hidden)
    }
```

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -34,11 +34,11 @@ resource "aws_vpc" "this" {
   assign_generated_ipv6_cidr_block = var.enable_ipv6
 
   tags = merge(
+    var.tags,
+    var.vpc_tags,
     {
       "Name" = format("%s", var.name)
     },
-    var.tags,
-    var.vpc_tags,
   )
 }
 
@@ -86,11 +86,11 @@ resource "aws_default_security_group" "this" {
   }
 
   tags = merge(
+    var.tags,
+    var.default_security_group_tags,
     {
       "Name" = format("%s", var.default_security_group_name)
     },
-    var.tags,
-    var.default_security_group_tags,
   )
 }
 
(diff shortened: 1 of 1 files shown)


antonbabenko:
Hi,

This is desired functionality. If a user does not want to specify `Name` tag, it will be set for them. Otherwise, tags will be overridden in the right sequence.

Also, this PR is not backward compatible because it will change tags for all existing users.

PR #721, opened 2021-12-27
docs: Add minimal example for S3 Gateway VPC Endpoint

Vlaaaaaaad:
## Description
This PR adds a minimal but complete example for an S3 Gateway VPC Endpoint. I would've loved to add a more complex example (ECR + S3), but the example was getting overwhelming.

## Motivation and Context
This PR is motivated by 2 reasons:
- the previous S3 example did not create a (free) Gateway Endpoint, which led to an embarrassing NAT Gateway bill for me
- it took me longer than I would've liked to figure out how to use the resulting VPC Endpoints

## Breaking Changes
N/A

## How Has This Been Tested?
I ran a `terraform fmt` and `terraform validate` on the `examples/complete-vpc` code. The VPC Endpoint definition was copied from a private project where everything seems to be working fine.

### DIFF
--- a/examples/complete-vpc/README.md
+++ b/examples/complete-vpc/README.md
@@ -44,6 +44,7 @@ Note that this example may create resources which can cost money (AWS Elastic IP
 |------|------|
 | [aws_iam_policy_document.dynamodb_endpoint_policy](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) | data source |
 | [aws_iam_policy_document.generic_endpoint_policy](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) | data source |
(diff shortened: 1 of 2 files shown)


bryantbiggs:
for DynamoDB and S3 there are two endpoint types: gateway and interface (the latter being the recent addition and what all other service endpoints are). In the example, we are showing that you can do either route for these specific two resources:
- S3 as an interface endpoint https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/5f5d8776687e403091a200901b2c452b3d3ad9b6/examples/complete-vpc/main.tf#L90
- DynamoDB as a gateway endpoint https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/5f5d8776687e403091a200901b2c452b3d3ad9b6/examples/complete-vpc/main.tf#L94

Instead of changing the example as you have done here, I would prefer to leave the example as is (to demonstrate/validate both use cases for these two specific services), but we can definitely improve on the documentation/notes to make it more clear to users

PR #737, opened 2022-01-25
fix: Addresses persistent diff with manage_default_network_acl

### LINKED ISSUE #736: Persistent diff when using `manage_default_network_acl` with subnets created outside config
## Description

When using `manage_default_network_acl = true`, creating any subnets outside of the terraform config results in a persistent diff.

## Versions

 - Terraform: 1.1.2

 - Provider(s): 3.73.0

 - Module: master branch

## Reproduction

1. cd examples/network-acls
2. run terraform apply
3. manually create a new subnet in the vpc
4. run terraform apply

## Expected behavior
Expected a clean terraform plan.

## Actual behavior

Terraform displays a persistent diff that cannot be reconciled.

```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # module.vpc.aws_default_network_acl.this[0] will be updated in-place
  ~ resource "aws_default_network_acl" "this" {
        id                     = "acl-0ec4aa76293d0d15f"
      ~ subnet_ids             = [
          - "subnet-075246afce2e34609",
            # (3 unchanged elements hidden)
        ]
        tags                   = {
            "Environment" = "dev"
            "Name"        = "network-acls-example"
            "Owner"       = "user"
        }
        # (5 unchanged attributes hidden)


        # (4 unchanged blocks hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

## Additional context

There is a prior issue reporting the same diff, but when _only_ using the module to manage subnets, https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/529.

That issue was resolved by adding all the subnets to the default nacl, https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/530.

However, that does not work when the subnet is created outside this module.

lorengordon:
As noted in the [terraform docs][0], subnets using the default network acl
will generate a persistent diff if they are not specified to the aws_default_network_acl
resource. This module was handling subnets created by the module, but
of course is not aware of subnets created externally to the module.

The docs suggest using lifecycle ignore_changes as an option to avoid
the persistent diff, which is the approach implemented in this patch.

Fixes #736

[0]: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_network_acl#managing-subnets-in-a-default-network-acl

## Breaking Changes

None

## How Has This Been Tested?

1. cd examples/network-acls
2. run terraform apply
3. manually create a new subnet in the vpc
4. run terraform apply
5. view that no changes are planned

### DIFF
--- a/examples/complete-vpc/main.tf
+++ b/examples/complete-vpc/main.tf
@@ -208,7 +208,7 @@ data "aws_iam_policy_document" "dynamodb_endpoint_policy" {
       test     = "StringNotEquals"
       variable = "aws:sourceVpce"
 
-      values = [data.vpc.vpc_id]
+      values = [module.vpc.vpc_id]
     }
   }
 }
--- a/examples/complete-vpc/main.tf
+++ b/examples/complete-vpc/main.tf
@@ -187,12 +187,6 @@ data "aws_security_group" "default" {
   vpc_id = module.vpc.vpc_id
 }
 
-# Data source used to avoid race condition
-data "aws_vpc_endpoint" "dynamodb" {
-  vpc_id       = module.vpc.vpc_id
-  service_name = "com.amazonaws.${local.region}.dynamodb"
-}
-
 data "aws_iam_policy_document" "dynamodb_endpoint_policy" {
   statement {
     effect    = "Deny"
--- a/examples/complete-vpc/README.md
+++ b/examples/complete-vpc/README.md
@@ -46,7 +46,6 @@ Note that this example may create resources which can cost money (AWS Elastic IP
 | [aws_iam_policy_document.dynamodb_endpoint_policy](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) | data source |
 | [aws_iam_policy_document.generic_endpoint_policy](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) | data source |
 | [aws_security_group.default](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/security_group) | data source |
-| [aws_vpc_endpoint.dynamodb](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/vpc_endpoint) | data source |
 
 ## Inputs
 
--- a/main.tf
+++ b/main.tf
@@ -587,28 +587,9 @@ resource "aws_default_network_acl" "this" {
 
   default_network_acl_id = aws_vpc.this[0].default_network_acl_id
 
-  # The value of subnet_ids should be any subnet IDs that are not set as subnet_ids
-  #   for any of the non-default network ACLs
-  subnet_ids = setsubtract(
-    compact(flatten([
-      aws_subnet.public[*].id,
-      aws_subnet.private[*].id,
-      aws_subnet.intra[*].id,
-      aws_subnet.database[*].id,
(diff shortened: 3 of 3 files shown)


antonbabenko (approved):
LGTM.

@bryantbiggs What do you think?

antonbabenko:
Could you also fix the issue introduced by #742 here? :)

lorengordon:
> Could you also fix the issue introduced by #742 here? :)

Haha, sure, on my phone, gimme 30 min too get back to a computer

lorengordon:
@antonbabenko Ok, should be good now!

lorengordon:
Also, not sure if you are tracking this already, but there is some kind of issue going on with the cidr_blocks outputs for unused features. After first apply, they are lists of nulls, but the second apply they are reported as a planned change, becoming lists of empty strings...

```
Changes to Outputs:
  ~ database_subnets_ipv6_cidr_blocks    = [
      - null,
      - null,
      - null,
      + "",
      + "",
      + "",
    ]
  ~ elasticache_subnets_ipv6_cidr_blocks = [
      - null,
      - null,
      - null,
      + "",
      + "",
      + "",
    ]
  ~ intra_subnets_ipv6_cidr_blocks       = [
      - null,
      - null,
      - null,
      + "",
      + "",
      + "",
    ]
  ~ private_subnets_ipv6_cidr_blocks     = [
      - null,
      - null,
      - null,
      + "",
      + "",
      + "",
    ]
  ~ public_subnets_ipv6_cidr_blocks      = [
      - null,
      - null,
      - null,
      + "",
      + "",
      + "",
    ]
  ~ redshift_subnets_ipv6_cidr_blocks    = [
      - null,
      - null,
      - null,
      + "",
      + "",
      + "",
    ]

You can apply this plan to save these new output values to the Terraform state, without changing any real infrastructure.
```

antonbabenko:
This PR is included in version 3.11.5 :tada:

antonbabenko:
Regarding the issue with outputs you are seeing, I don't think we have seen such behavior before.

lorengordon:
> Regarding the issue with outputs you are seeing, I don't think we have seen such behavior before.

Roger, I'll create a separate issue for that and see if I can track it down.

examples/complete-vpc/main.tf
```diff
@@ -208,7 +202,7 @@ data "aws_iam_policy_document" "dynamodb_endpoint_policy" {
       test     = "StringNotEquals"
       variable = "aws:sourceVpce"
 
-      values = [data.vpc.vpc_id]
+      values = [module.vpc.vpc_id]
```
bryantbiggs, line 205: oy, this is still wrong! it should be

```suggestion
      values = [data.aws_vpc_endpoint.dynamodb.id]
```

because the condition is "source VPC endpoint" not "source VPC"

lorengordon: That's what it was before #742 (attempted to) change it to the vpc id! Oi! Revert and revert and revert!

bryantbiggs: we can leave it as is now, but the condition is wrong - should be `"aws:SourceVpc"`

lorengordon: Looking at it again now, the data source *is* a race condition. The comment makes no sense. This whole thing is confusing to me. Why not just get the vpce id from an output of the vpc endpoint module?

Issue #740, opened 2022-01-26
Transit Gateway private route support


henryk-radoslaw-rychlik:
Hello,

I'm using this module to set up 2 different VPCs and tie them together only on private side with a Transit Gateway.
The module supports passing custom routes for the main route table via the following options:
```
...
  default_route_table_routes = [
    {
      cidr_block        = "10.191.0.0/16"
      transit_gateway_id = module.tgw.ec2_transit_gateway_id
    }
  ]
  manage_default_route_table = true
...
}
```

This would suffice, if the private subnets didn't have explicitly associated route tables. I don't see a documented way of passing an additional route to the private or public route tables.

Here are the changes made locally and tested that you can use for a PR:
`main.tf`
```
resource "aws_route" "private_route_table_routes" {
  count = length(var.private_subnets) * length(var.private_route_table_routes)

  route_table_id         = element(aws_route_table.private[*].id, floor(count.index/length(var.private_route_table_routes)))
  destination_cidr_block = element(var.private_route_table_routes, floor(count.index/length(var.private_subnets))).cidr_block

  # One of the following targets must be provided
  egress_only_gateway_id = lookup(element(var.private_route_table_routes, floor(count.index/length(var.private_subnets))), "egress_only_gateway_id", null)
  gateway_id             = lookup(element(var.private_route_table_routes, floor(count.index/length(var.private_subnets))), "gateway_id", null)
  instance_id            = lookup(element(var.private_route_table_routes, floor(count.index/length(var.private_subnets))), "instance_id", null)
  nat_gateway_id         = lookup(element(var.private_route_table_routes, floor(count.index/length(var.private_subnets))), "nat_gateway_id", null)
  network_interface_id   = lookup(element(var.private_route_table_routes, floor(count.index/length(var.private_subnets))), "network_interface_id", null)
  transit_gateway_id     = lookup(element(var.private_route_table_routes, floor(count.index/length(var.private_subnets))), "transit_gateway_id", null)
  vpc_endpoint_id        = lookup(element(var.private_route_table_routes, floor(count.index/length(var.private_subnets))), "vpc_endpoint_id", null)
  vpc_peering_connection_id = lookup(element(var.private_route_table_routes, floor(count.index/length(var.private_subnets))), "vpc_peering_connection_id", null)

  timeouts {
    create = "5m"
  }
}
```

`variables.tf`
```
variable "private_route_table_routes" {
  description = "Configuration block of private routes."
[...]

PR #741, opened 2022-01-26
Added support for custom private routes

henryk-radoslaw-rychlik:
Tested, described reasoning in the following issue:
https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/740


khdevel:
Hello, what is the status here? That would be good to have such feature in this module.

antonbabenko:
@khdevel We will resume reviewing and accepting PRs like this one a bit later. We are busy with other priorities (e.g. updating code for AWS provider v4) at the moment.

PR #742, opened 2022-01-27
chore: Fixed code in examples (dynamodb resource not found error)

tomahawk-pilot:
fix for dynamodb resource not found error in the example.

### DIFF
--- a/examples/complete-vpc/main.tf
+++ b/examples/complete-vpc/main.tf
@@ -208,7 +208,7 @@ data "aws_iam_policy_document" "dynamodb_endpoint_policy" {
       test     = "StringNotEquals"
       variable = "aws:sourceVpce"
 
-      values = [data.aws_vpc_endpoint.dynamodb.id]
+      values = [data.vpc.vpc_id]
     }
   }
 }


lorengordon:
@tomahawk-pilot @antonbabenko This change is not valid... There is no data source named "vpc"... This change is causing the pre-commit failures in the pr I opened here, https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/737

lorengordon:
(And yes, I am also plugging my PR to try to get review eyes on it 😁)

tomahawk-pilot:
> @tomahawk-pilot @antonbabenko This change is not valid... There is no data source named "vpc"... This change is causing the pre-commit failures in the pr I opened here, https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/737

I ran the tf and it works. But on second look it should be module.vpc

lorengordon:
I suppose it would be good if changes to the examples also ran all the precommit jobs, to catch these things earlier?

bryantbiggs:
all code in the module repos are checked by pre-commit - but this is a logic error. you could have just as easily put "foo" instead, the policy doesn't care what value is there - it just gets evaluated as allow or deny

lorengordon:
Ok... Please look at the PR I linked. Clearly many more tests were run on that PR, and two of them caught this failure.

bryantbiggs:
Hmm, actually this one should have been caught. It's syntactically incorrect

antonbabenko:
This PR is included in version 3.11.5 :tada:

bryantbiggs:
So I see the issue here which has caused a flurry of spurious changes. The DynamoDB endpoint is a gateway endpoint and I guess that is not being found by the data source. We can swap this for one of the interface endpoints like SSM

__ OR __

We change the condition to be `"aws:SourceVpc"` (remove trailing `e` so its for the VPC and not the VPC endpoint)

PR #744, opened 2022-01-28
feat: Dedicated subnets for Transit Gateway Attachement

osaluden:
## Description
Add a new subnet type : tgwattach. The tgwattach subnet are usually small subnet range with no route table associated with unless you set the parameter  `tgwattach_associate_private_route_table` to true to benefit from the default route to NAT Gateway in case of an outbound VPC.

## Motivation and Context
Following AWS best practices for Transit Gateway attachement: https://docs.aws.amazon.com/vpc/latest/tgw/tgw-best-design-practices.html:
*Use a separate subnet for each transit gateway VPC attachment. For each subnet, use a small CIDR, for example /28, so that you have more addresses for EC2 resources*

These attachement subnets does not need ACLs neither route table, unless the VPC to be attached is an outbound VPC. In this case a route to NAT Gateway must be set. This is done by associating the `private` route table to those subnets with the parameter `tgwattach_associate_private_route_table` (set it to `true`or `false`)

(if ACL is required, it is possible to enhance the module and add ACLs for the tgwattach subnets)

## Breaking Changes
No breaking change noticed.

## How Has This Been Tested?
I tested the changes with a hub and spoke architecture, using a transit Gateway, three `application VPCs` (no internet gateway and NAT Gateway) and an outbound VPC (with IGW and NAT GW)
The new module is compatible with the Transit Gateway module. Attachement is performed thanks to the `tgwattach_subnets` IDs output.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -4,6 +4,7 @@ locals {
     length(var.elasticache_subnets),
     length(var.database_subnets),
     length(var.redshift_subnets),
+    length(var.tgwattach_subnets),
   )
   nat_gateway_count = var.single_nat_gateway ? 1 : var.one_nat_gateway_per_az ? length(var.azs) : local.max_subnet_length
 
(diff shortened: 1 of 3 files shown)


osaluden:
Hi there
Do you have any news on this PR? Does it make sense to you?
Thank you for your feedbacks / advices.

madpellzo:
Hi there
Do you have any news regarding this PR? Any progress?
Thank you for you kind help.

osaluden:
Hi there
Do you have any news regarding this PR? Any progress?
Thank you for you kind help.

PR #746, opened 2022-02-03
feat: Add full subnet resource output for each subnet type

patrick-vinas-ripcord:
## Description
Adds as outputs maps of AZ to subnet ID for each subnet type (private, public, database, etc)

## Motivation and Context
Having a mapping of which subnets were created in each AZ is useful e.g. when creating EKS node groups for each AZ with the terraform EKS module. This output would allow creation of per-az node groups with the AZ in the node group name.

## Breaking Changes
Does not break backward compatibility with anything, it is just a collection of new outputs.

## How Has This Been Tested?
- [ ] Verified that the output works with no subnets and also multiple subnets

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -68,6 +68,11 @@ output "vpc_owner_id" {
   value       = try(aws_vpc.this[0].owner_id, "")
 }
 
+output "private_subnet_resources" {
+  description = "List of private subnet resources"
+  value       = aws_subnet.private
+}
+
 output "private_subnets" {
   description = "List of IDs of private subnets"
   value       = aws_subnet.private[*].id
@@ -76,11 +81,6 @@ output "private_subnets" {
 output "private_subnet_arns" {
   description = "List of ARNs of private subnets"
   value       = aws_subnet.private[*].arn
-}
-
-output "private_subnet_azs" {
-  description = "Map of AZs to private subnet IDs"
-  value       = { for subnet in aws_subnet.private[*] : subnet.availability_zone => subnet.id... }
 }
 
 output "private_subnets_cidr_blocks" {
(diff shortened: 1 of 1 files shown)


outputs.tf
```diff
@@ -78,6 +78,11 @@ output "private_subnet_arns" {
   value       = aws_subnet.private[*].arn
 }
 
+output "private_subnet_azs" {
+  description = "Map of AZs to private subnet IDs"
+  value       = { for subnet in aws_subnet.private[*] : subnet.availability_zone => subnet.id... }
```
bryantbiggs, line 83: Is this a typo, or new splat syntax I haven't seen 😅 
```suggestion
  value       = { for subnet in aws_subnet.private[*] : subnet.availability_zone => subnet.id }
```

patrick-vinas-ripcord: The ellipsis enables grouping by key, so you end up with a map like: { us-east-1 => [subnet-1, subnet-2, etc] }

bryantbiggs: I'd also be more keen to just output something like the following and users can slice and dice, pick and choose how they see fit

```hcl
output "private_subnets" {
  description = "Map private subnets created and their attributes"
  value       = aws_subnet.private
```

but the name conflicts with an output we have already

patrick-vinas-ripcord: I removed the map output and added a full resource output, and updated the title of the PR

antonbabenko:
I don't think this change is good for several reasons like:
1. It will periodically produce errors like "Objects have changed outside of Terraform" (related bug in Terraform -  https://github.com/hashicorp/terraform/issues/28803 ) when something is managed outside of this module (e.g. routes, some attachments, etc).
2. It increases the output significantly.

The most stable approach is to use an extra data source like [aws_subnet](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/subnet), `aws_subnets` or `aws_subnet_ids` to get the required attributes of the subnets in places where you need them.

Closing this PR.

PR #773, opened 2022-03-31
feat: Change to allow create variable within specific vpc objects

matipadeveloper:
## Description
This change is to allow for create to work within each vpc endpoint object. If the create is outside of a vpc endpoint then that will take priorty over the specific vpc endpoints create.

## Motivation and Context
This will make the script alot more flexible and fine grain. As you can go down to object level to specify which objects should be created or not. Making a lot more use cases possible.

## Breaking Changes
The only thing that was introduced that would be debtable is try but it seems try has been there for a long time so seem like it is backwards compatible. 

## How Has This Been Tested?
Yes I have an example project which I tested this on. I tested with multiple use cases to make sure.
This change will also default create to true with object level vpc endpoints just like it has been. Meaning create is still optional for object level and resource level create.

### DIFF
--- a/examples/complete-vpc/main.tf
+++ b/examples/complete-vpc/main.tf
@@ -121,7 +121,7 @@ module "vpc_endpoints" {
       subnet_ids          = module.vpc.private_subnets
     },
     ecs_telemetry = {
-      create = false
+      create              = false
       service             = "ecs-telemetry"
       private_dns_enabled = true
       subnet_ids          = module.vpc.private_subnets
--- a/examples/complete-vpc/main.tf
+++ b/examples/complete-vpc/main.tf
@@ -89,7 +89,6 @@ module "vpc_endpoints" {
 
   endpoints = {
     s3 = {
-      create = false
       service = "s3"
       tags    = { Name = "s3-vpc-endpoint" }
     },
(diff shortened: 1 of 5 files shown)


antonbabenko:
Looks good. Could you please update the code in `examples/complete-vpc` and include one line with `create = false` to show this feature?

In this block, for example:
https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/e02118633f268ff1f86021a8fa9f3afcd1c37d85/examples/complete-vpc/main.tf#L123-L127

matipadeveloper:
> Looks good. Could you please update the code in `examples/complete-vpc` and include one line with `create = false` to show this feature?
> 
> In this block, for example:
> 
> https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/e02118633f268ff1f86021a8fa9f3afcd1c37d85/examples/complete-vpc/main.tf#L123-L127

Awesome I have added it in the complete vpc

examples/complete-vpc/main.tf
```diff
@@ -121,6 +121,7 @@ module "vpc_endpoints" {
       subnet_ids          = module.vpc.private_subnets
     },
     ecs_telemetry = {
+      create = false
```
antonbabenko, line 124:
```suggestion
      create              = false
```

antonbabenko:
This PR is included in version 3.14.0 :tada:

antonbabenko:
Thank you @matipadeveloper for this PR!

matipadeveloper:
Anytime @antonbabenko thank you for the module!

PR #786, opened 2022-05-18
fix: Name key of vpc tags is overriding Name key cgw

avsprashant:
If you specify Name key in vpc tags, then Name key of vpc tags is overriding Name key of cgw

## How Has This Been Tested?


antonbabenko:
This has been done on purpose. The ordering is correct. `Name` is a special tag.

Issue #790, opened 2022-05-29
Add output for flow_log_log_group_name_prefix


kderck:
## Is your request related to a new offering from AWS?

 - No 🛑

## Is your request related to a problem? Please describe.

`flow_log_cloudwatch_log_group_name_prefix` is not output which means you can not create a metric filter with the outputs available. The two outputs (`vpc_flow_log_id`, `vpc_flow_log_destination_arn` ) can not be used by a metric filter as part of the `log_group_name` argument

## Describe the solution you'd like.

I would like to have `flow_log_cloudwatch_log_group_name_prefix` output as a output. This will allow for a metric filter to be created for the relevant vpc by using the vpc id and prefix of the log group. 

## Describe alternatives you've considered.

You could use a `data` block to look up the `log_group_name` from `flow_log_destination_arn`. However this Introduces more boilerplate code and the need for a lookup, and can complicate scenarios where you may be creating multiple vpcs with `for_each`. Unlike resources which all arguments are exported as attributes, modules use inputs which need to be explicitly added as a output. Allowing us to simply retrieve the flow_log prefix by using a module output, and by using the already existing vpc_id you can get the CloudWatch Log Group name by interpolation.

## Additional context

[Terraform Documentation: CloudWatch Metric Filter](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_log_metric_filter)

kderck:
On further research. It isn't possible to use a `data` block as `aws_cloudwatch_log_group` doesn't support a `arn` lookup.

antonbabenko:
Answered in https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/792#issuecomment-1140410077

PR #792, opened 2022-05-29
feat: Add flow_log_log_group_name_prefix as a output

kderck:
## Description

Adds `flow_log_log_group_name_prefix` as an output.

## Motivation and Context

See [#790](https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/790)

## Breaking Changes

No.

## How Has This Been Tested?


antonbabenko:
If you can pass variables to a module, you should be able to use that value without using module's outputs.

This change is not necessary.

kderck:
I disagree, when you use a `variable` terraform creates an graph it's not guaranteed to map the dependencies correctly as if the said variable was an output of the resource that creates it `terraform` will need to wait for the apply therefore you need the `depends_on` argument. However, in this scenario I have followed convention.

Issue #803, opened 2022-06-22
VPC Endpoint sgroup remove loop with no security group defined (security_group_ids)


Roxyrob:
## Description
In `aws_vpc_endpoint` terraform resource as per documentation: `If no security groups are specified, the VPC's [default security group]... is associated with the endpoint.`

This vpc endpoint module correctly creates vpc endpoint and associate vpc defaul security group on first apply , but on subsequent applies want to remove associated default security group.

## Versions

- Module version [Required]: v3.14.2

- Terraform version: v1.2.2

- Provider version(s): registry.terraform.io/hashicorp/aws v4.12.0

## Reproduction Code [Required]

```
module "vpce" {
  source  = "terraform-aws-modules/vpc/aws//modules/vpc-endpoints"
  version = "v3.14.2"

  vpc_id     = "vpc-..."
  endpoints  = {
    s3test1 = {
      service = "s3"
    },
  }
}
```

terraform apply
   => first time vpc endpoint will be correctly created (with vpc default security group associated)
   => all subsequent times returns:

<pre>
Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:
  # module.vpce.aws_vpc_endpoint.this["s3test1"] will be updated in-place
  ~ resource "aws_vpc_endpoint" "this" {
        id                    = "vpce-..."
      <b>~ security_group_ids    = [
          - "sg-...",
        ]</b>
      ...
    }
Plan: 0 to add, <b>1 to change</b>, 0 to destroy.
</pre>

## Expected behavior

With no security group ids defined default vpc security group should remain associated with vpc endpoint

## Actual behavior

With no security group ids defined default vpc security group is removed by vpc endpoint configuration

PR #806, opened 2022-07-03
fix: Fix for default security group loop

matt-empson:
## Description
This change introduces a data source to look up the default security group of the vpc that the endpoint will be deployed into. The default security group is then explicitly associated with the endpoint if no other security groups are passed into the module, or found by the lookup function. This resolves the issue of terraform wanting to make changes on every apply, as described in the linked issue below.

## Motivation and Context
This is to fix what's described in https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/803

## Breaking Changes
No breaking changes

## How Has This Been Tested?

The vpc in the complete example was deployed prior to making my change. Terraform apply was run again after making my change, and terraform detected no changes to make.

Locally, I also deployed the simple-vpc example with the below endpoint configuration added to it

```
module "vpc_endpoints" {
  source = "../../modules/vpc-endpoints"

  vpc_id = module.vpc.vpc_id

  endpoints = {
    s3 = {
      service = "s3"
      tags    = { Name = "s3-vpc-endpoint" }
    }
  }

  tags = {
    Owner       = "user"
    Environment = "dev"
  }
}
```

I ran terraform apply again and was able to reproduce the problem described in the linked issue. After adding my change and running a terraform apply, no changes were detected, which is the expected behaviour.

### DIFF
--- a/modules/vpc-endpoints/README.md
+++ b/modules/vpc-endpoints/README.md
@@ -73,6 +73,7 @@ No modules.
 | Name | Type |
 |------|------|
 | [aws_vpc_endpoint.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_endpoint) | resource |
+| [aws_security_group.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/security_group) | data source |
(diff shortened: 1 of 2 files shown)


matt-empson:
Still current.

matt-empson:
Still current.

PR #808, opened 2022-07-08
Allow additional routes for pub, intra, priv tables

bogdando:
Allow additional routes to be added for the public, private and intra
routing tables, like it is already available for the default table.

An example use case could be peering with some VPC and making it routed
with the VPC managed by the module. This example assumes that
requesting and accepting the peering request should be handled by a
user. As well as adding default routes on the other (peered VPC) side.

Signed-off-by: Bogdan Dobrelia <bogdan.dobrelia@spearline.com>

## How Has This Been Tested?


bogdando:
Oops, sorry, I was planning to only draft it

PR #809, opened 2022-07-08
feat: Allow additional routes for intra and private routing tables

bogdando:
## Description

Allow additional routes to be added for the private and intra
routing tables, like it is already available for the default table.

An example use case could be peering with some VPC and making it routed
with the VPC managed by the module. This example assumes that
requesting and accepting the peering request should be handled by a
user. As well as adding default routes on the other (peered VPC) side.

## Motivation and Context

The use case is configuring all required routes within the VPC module, instead of adding custom
calls for ``aws_route``. The latter is not always possible with terragrunt modules for example, which allows
only other modules there, and no custom resources.
(see https://github.com/gruntwork-io/terragrunt/issues/1187 and https://github.com/gruntwork-io/terragrunt/issues/1150)

Custom routes may be required for the private/intra VPC route tables to cover many cases, like
peering the managed VPC with another VPC (outside of this feature scope) and manage additional
routes on both sides. This change allows custom routes to be added for the managed VPC, for this
example. While adding the needed routes on the other side of the peering, and creating the peering would be
left for the user exercise.

## How Has This Been Tested?
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects (with my [tEKS fork](https://github.com/bogdando/teks/blob/main/QUICKSTART.md#demo-dev-cluster-un-this-fork))


samsen1:
What's the hold up w/ this? We actually need this for transit gateway attachment

README.md
```diff
@@ -342,6 +342,8 @@ No modules.
 | <a name="input_default_route_table_propagating_vgws"></a> [default\_route\_table\_propagating\_vgws](#input\_default\_route\_table\_propagating\_vgws) | List of virtual gateways for propagation | `list(string)` | `[]` | no |
 | <a name="input_default_route_table_routes"></a> [default\_route\_table\_routes](#input\_default\_route\_table\_routes) | Configuration block of routes. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_route_table#route | `list(map(string))` | `[]` | no |
 | <a name="input_default_route_table_tags"></a> [default\_route\_table\_tags](#input\_default\_route\_table\_tags) | Additional tags for the default route table | `map(string)` | `{}` | no |
+| <a name="input_private_route_table_routes_extra"></a> [default\_route\_table\_routes](#input\_default\_route\_table\_routes) | Configuration block of additional routes for private subnets route tables. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_route_table#route | `list(map(string))` | `[]` | no |
```
lmpardey, line 345: For accuracy it would be better to use the link to the `aws_route_table` resource instead of `aws_default_route_table`, since this change is affecting resources of the former type. Link is very similar: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route_table#route

README.md
```diff
@@ -342,6 +342,8 @@ No modules.
...
+| <a name="input_intra_route_table_routes_extra"></a> [default\_route\_table\_routes](#input\_default\_route\_table\_routes) | Configuration block of additional routes for intra subnets route tables. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_route_table#route | `list(map(string))` | `[]` | no |
```
lmpardey, line 346: Same here, For accuracy it would be better to use the link to the `aws_route_table` resource instead of `aws_default_route_table`, since this change is affecting resources of the former type. Link is very similar: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route_table#route

main.tf
```diff
@@ -228,6 +228,25 @@ resource "aws_route_table" "private" {
 
   vpc_id = local.vpc_id
 
+  dynamic "route" {
```
lmpardey, line 231: Unfortunately, adding the `route` element here will conflict with the rules that are created if a NAT gateway is also created. See the warning at the top of this page: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route_table
> NOTE on Route Tables and Routes:
>
> Terraform currently provides both a standalone [Route resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route) and a Route Table resource with routes defined in-line. At this time you cannot use a Route Table with in-line routes in conjunction with any Route resources. Doing so will cause a conflict of rule settings and will overwrite rules.

In order to support both NAT gateway routes on the private route tables, and the ones provided via the new parameter, you would need to use an `aws_route` resource and `for_each` over the private route tables while also applying the change to every route table created (one per nat gateway).

For example, it would need something like this (illustrative example only):

```
resource "aws_route" "private_route" {
  # setproduct of two sets gives a pair, so each.value[0] is the route table ID and
  # each.value[1] is a mapping of route table values.
  for_each = setproduct(aws_route_table.private[*].id, var.private_route_table_routes_extra)
  
  route_table_id = each.value[0]  # refers to aws_route_table.private[*].id

  destination_cidr_block = each.value[1].destination_cidr_block
  ipv6_cidr_block = lookup(each.value[1], "destination_ipv6_cidr_block", null)
  ... # and so on
}
```

variables.tf
```diff
@@ -418,6 +418,18 @@ variable "default_route_table_tags" {
...
 }
 
+variable "private_route_table_routes_extra" {
+  description = "Configuration block of additional routes for private subnets route tables. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_route_table#route"
+  type        = list(map(string))
+  default     = []
+}
+
+variable "intra_route_table_routes_extra" {
+  description = "Configuration block of additional routes for intra subnets route tables. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_route_table#route"
```
lmpardey, line 428: The link on this variable should reflect the resource which it would affect, so https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route_table#route.

lmpardey (changes requested):
Thanks for the contribution! There are some cosmetic issues with the documentation, and one significant issue with the private route table, due to the way Terraform handles the combination of inline and standalone routes. If not resolved, this could cause issues when deploying a VPC with NAT gateway and custom private routes.

bogdando:
Just a heads up, I didn't forget about this one, will come back to it once I have spare time to rework it for NAT GW support :)

bogdando:
@lmpardey could you please reopen it?
I'm preparing a fix, as you recommended and update this PR shortly.

It is almost done, just that the code that works in console, fails when being applied
works in console:
```
> setproduct(tolist(aws_route_table.private[*].id),[for route in [split(",", "172.30.0.0/16,192.168.142.0/24")]:{"vpc_peering_connection_id" = "foo","cidr_block" = route}])
tolist([
  [
    "rtb-02a1877c819b737d6",
    {
      "cidr_block" = tolist([
        "172.30.0.0/16",
        "192.168.142.0/24",
      ])
      "vpc_peering_connection_id" = "foo"
    },
  ],
  [
    "rtb-0484ba914f31f866a",
    {
      "cidr_block" = tolist([
        "172.30.0.0/16",
        "192.168.142.0/24",
      ])
      "vpc_peering_connection_id" = "foo"
    },
  ],
])
```
fails if applied:
```
│ Error: Invalid for_each set argument
│ 
│   on main.tf line 1047, in resource "aws_route" "private_route_custom":
│ 1047:   for_each = toset(setproduct(aws_route_table.private[*].id, var.private_routes_extra))
│     ├────────────────
│     │ aws_route_table.private is tuple with 2 elements
│     │ var.private_routes_extra is list of map of string with 2 elements
│ 
│ The given "for_each" argument value is unsuitable: "for_each" supports maps
│ and sets of strings, but you have provided a set containing type tuple.
╵
``` :(

bogdando:
Re-submitted as a new one https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/833
PTAL

PR #823, opened 2022-08-23
feat: Add transit gateway vpc attachment feature

saravanan30erd:
## Description
Transit gateway is used to connect many vpcs from same and across many accounts. So Transit gateway creation can be different module but we need a way to attach the newly created VPCs to Transit Gateway.

## Motivation and Context
Here I am attaching the VPC to given Transit Gateway when `attach_tgw` is `true`. Also creating routes in private route table when `tgw_destination_cidrs` is provided. I chose private subnets(its route tables) as default for attaching with Transit Gateway because public subnets mostly used for proxy/loadbalancer and private subnets mostly used for running applications (those applications might require to communicate other resources in another VPC internally via transit gateway).

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,16 @@ Sometimes it is handy to have public access to Redshift clusters (for example if
   enable_public_redshift = true  # <= By default Redshift subnets will be associated with the private route table
 ```
 
+## Transit Gateway (TGW) Attachment
+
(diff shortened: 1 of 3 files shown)


bodgit:
The [Transit Gateway best practices](https://docs.aws.amazon.com/vpc/latest/tgw/tgw-best-design-practices.html) recommend creating dedicated /28 subnets for VPC attachments. I was initially attaching to the private subnets but switched over to using some dedicated subnets.

Issue #820, opened 2022-08-03
Customize Private and Public Route Table


cabewill:
## Is your request related to a new offering from AWS?
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - No 🛑: please wait to file a request until the functionality is avaialble in the AWS provider

## Is your request related to a problem? Please describe.

When I create a VPC a need to add more routes in my route table so I can conncet to peering connections or to other services. I can't do this with this module.

## Describe the solution you'd like.

There is already a code to do this with the default root table. It's necessary to replicate the same solution to others root tables.

## Describe alternatives you've considered.

I have to create a new resourte aws_route just to add new routes to the recently created route tables

pickgr:
I also recently ran into this limitation where I needed to update the route table for a VPC to VPC peering connection.

If the route tables where created without the inline route and instead used

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route

then this would be possible.

Per the documentation:

> NOTE on Route Tables and Routes:
Terraform currently provides both a standalone Route resource and a [Route Table](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route_table) resource with routes defined in-line. At this time you cannot use a Route Table with in-line routes in conjunction with any Route resources. Doing so will cause a conflict of rule settings and will overwrite rules.

samsen1:
This PR is staled but it may be useful to someone?

https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/809

bogdando:
I hope I can move it forward shortly

bogdando:
The above change request didn't make it through review, unfortunately, Another way forward for Terragrunt users would be adding a terraform registry module for aws_route (where should I open a ticket for it, @antonbabenko ?). In Terraform one can just compose additional ``aws_route`` resources as needed, but not in ``terragrunt.hcl`` files, without such a module URI to refer.

Once there is an ``aws_route`` module in [terraform registry](https://registry.terraform.io/browse/modules) ``terragrunt.hcl`` files could refer it via [tfr](https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/#terraform) transport, or a github (https transport) ``*/modules/aws_route`` URI.

[...]

PR #833, opened 2022-09-27
feat: Allow additional routes for intra and private routing tables

bogdando:
## Description

Allow additional routes to be added for the private and intra routing tables, like it is already available for the default table.

An example use case could be peering with some VPC and making it routed with the VPC managed by the module. This example assumes that requesting and accepting the peering request should be handled by a user. As well as adding default routes on the other (peered VPC) side.

Ensure no conflicts for additional routing rules with private routes defined for NAT gateway managed with this module.

In order to support both NAT gateway routes on the private route tables, and the ones provided via the new parameters for custom routes, use an aws_route resource instead of the one for managing route tables. That is, per the documentation:

    NOTE on Route Tables and Routes: Terraform currently provides both
    a standalone Route resource and a Route Table resource with routes
    defined in-line. At this time you cannot use a Route Table with
    in-line routes in conjunction with any Route resources. Doing so
    will cause a conflict of rule settings and will overwrite rules.

To keep input values for custom default route table, private and intra routing rules consistent and interchangeble:
 * use the same aws_route approach for both private and intra cases;
 * alias (ipv6_)cidr_block and destination_(ipv6_)cidr_block so that both inputs can be used interchangebly for default route table and private/intra subnet route resources.

## Motivation and Context

The use case is configuring all required routes within the VPC module, instead of adding custom
calls for ``aws_route``. The latter is not always possible with terragrunt modules for example, which allows
only other modules there, and no custom resources.
(see https://github.com/gruntwork-io/terragrunt/issues/1187 and https://github.com/gruntwork-io/terragrunt/issues/1150)

Custom routes may be required for the private/intra VPC route tables to cover many cases, like
peering the managed VPC with another VPC (outside of this feature scope) and manage additional
routes on both sides. This change allows custom routes to be added for the managed VPC, for this
example. While adding the needed routes on the other side of the peering, and creating the peering would be
left for the user exercise.

Implements https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/820

## How Has This Been Tested?
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects (with my [tEKS fork](https://github.com/bogdando/teks/blob/main/QUICKSTART.md#demo-dev-cluster-un-this-fork))


main.tf
```diff
@@ -1041,6 +1041,77 @@ resource "aws_nat_gateway" "this" {
...
+  # consume it as end values. Iterate over a map with keys names matching
+  # router IDs (key unquieness is ensured by a hash of each rule)
+  # as for_each cannot take these from a list.
+  for_each = {
+    for elements in chunklist(flatten([
+      for pair in setproduct(
+        aws_route_table.private[*].id,
+        [var.private_routes_extra]
+      ) : setproduct([pair[0]], [for p in pair[1] : [jsonencode(p)]])
+    ]), 2) : "${elements[0]}:${sha256(elements[1])}" => jsondecode(elements[1])
```
antonbabenko, line 1056: Sorry, but this code is not what we want to accept in this module...

antonbabenko (commented):
We expect users to manage many other network-related resources outside of this module, even if such resources belong to VPC because there are a lot of variations and use cases.

We won't accept this PR because it is too complicated.

We may add support for `aws_route` resources in future major releases.

main.tf
bogdando: Thank you for feedback. I couldn't find another approach to make setproduct and for_each working on different data types - strings for route tables IDs and lists of maps for roles. If you have ideas to share, I'd appreciate. I'm quite new to terraform, so please excuse my ignorance.

bogdando: Please also note linked terragrunt issues,. It seems there is no github URL to use aws_route as a dedicated module alongside vpc one

bogdando:
I workarounded this in my teks fork by using [another module](https://github.com/bogdando/teks/commit/be2398422497d80c9c6d7340c2799967fb00cb98) to manage VPC peering and custom routes

Issue #839, opened 2022-10-11
Default CIDR for module should be from RFC 1918


air3ijai:
## Description

Accordingly to the module input description, this module does not require any inputs because all of them have default values.

[Optional Inputs](https://registry.terraform.io/modules/terraform-aws-modules/vpc/aws/latest?tab=inputs#optional-inputs)

> These variables have default values and don't have to be set to use this module. You may set these variables to override their default values. This module has no required variables.

It means that to use module we shoul be abble to use a simple code:
```javascript
module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = "3.16.0"
}
```

But this doesn't work :)

## Versions

- Module version: `3.16.0`
- Terraform version: `Terraform v1.3.1 on darwin_arm6`
- Provider version(s): `hashicorp/aws v4.34.0`

## Reproduction code

```javascript
module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = "3.16.0"
}
```

 1. Authenticate on AWS
 2. `vi main.tf`
 3. Add sample code from the block above
 4. `terraform init`
 5. `terraform apply`

## Expected behavior

Code should be run as described in documentation

## Actual behavior

We get an error with the message to set the `cidr_block`
```
╷
│ Error: expected "cidr_block" to contain a network Value with between 16 and 28 significant bits, got: 0
│
│   with module.vpc.aws_vpc.this[0],
│   on .terraform/modules/vpc/main.tf line 23, in resource "aws_vpc" "this":
│   23:   cidr_block          = var.use_ipam_pool ? null : var.cidr
│
╵
```

### Terminal Output Screenshot(s)

<img width="771" alt="Screen Shot 2022-10-11 at 20 15 55" src="https://user-images.githubusercontent.com/88528265/195159107-550f8a0f-dbdb-43c0-81f7-0fab83bc1496.png">

## Additional context

Default value of the `cidr_block` is `0.0.0.0/0` - https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/variables.tf#L13,L17

This value doesn't look to be the correct one and in case of VPC, for most cases, it is intended to be for private use and we should use blocks from RFC 1918 range.

Accordingly to the AWS documentation - [IPv4 VPC CIDR blocks](https://docs.aws.amazon.com/vpc/latest/userguide/configure-your-vpc.html#vpc-cidr-blocks)
[...]

PR #842, opened 2022-10-13
feat: Use default `cidr` value from RFC 1918 (#839)

air3ijai:
## Description
We should use default value from RFC 1918 to simplify module usage without specifying additionally `cidr`

## Motivation and Context
This is related to the [Default CIDR for module should be from RFC 1918 #839](839)

## Breaking Changes
No breaking changes

## How Has This Been Tested?


antonbabenko:
`10.0.0.0/16` is an example CIDR mentioned by AWS, but not as a default or recommended.

PR #849, opened 2022-10-25
feat: Network ACL

SlavaSubotskiy:
## Description
A dedicated `network-acl` sub-module was added.

## Motivation and Context
We are using `terragrunt` to describe our infrastructure.
We have standardized subnet naming (e.g. "Transfer", "Transit", etc.) and subnets already provisioned for us by another team.
All we need to do is to create and associate NACLs with them.
The current VPC module allows us to create NACLs, but only the predefined ones (e.g. "public", "outpost", "redshift", etc.).
To support custom NACLs a simple module was created.
It has an optional `subnet_ids` input to associate NACL with subnets.
We are using this module within the organization and now decided to contribute it to the public VPC module.

## Breaking Changes
No

## How Has This Been Tested?

`examples/network-acl` was used for the testing with existing VPC and subnet


antonbabenko:
I completely understand what you want to achieve, but I don't think we should be maintaining it as a separate submodule for two reasons:
1. We need to improve the module to use some of the later features of Terraform (like we have done in EKS, RDS, and some other modules already).
2. If we add a NACL submodule, we need to add one for subnets, routes, routing tables, NAT gateways, etc. This is too much repetitive work.

I propose that instead of doing a separate module, we look into extending this module by managing extra resources (like NACL, subnets, etc) **in the provided VPC**. This is doable, but only after the first step from above is implemented (specifically, by using `for_each` instead of `count`).

@bryantbiggs WDYT?

tcharewicz:
> I propose that instead of doing a separate module, we look into extending this module by managing extra resources (like NACL, subnets, etc) **in the provided VPC**. This is doable, but only after the first step from above is implemented (specifically, by using `for_each` instead of `count`).
> 
> @bryantbiggs WDYT?

For me it's sound great, How can I help to refactor this module?

My example sub-modules with for each loops.
https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/834/files

PR #853, opened 2022-10-31
feat: Add separate EKS subnets

lordz-md:
## Description
Added EKS subnets

## Motivation and Context
Need to have separate EKS and Private network and be able to enable NAT on EKS.

## Breaking Changes
No breaking changes.

## How Has This Been Tested?
Ran pre-commit.
Created a local VPC with required changes,

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,6 +1,7 @@
 locals {
   max_subnet_length = max(
     length(var.private_subnets),
+    length(var.eks_subnets),
     length(var.elasticache_subnets),
     length(var.database_subnets),
     length(var.redshift_subnets),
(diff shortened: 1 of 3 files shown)


lordz-md:
@antonbabenko, after your latest updates we found the need to be able to have separate private and EKS subnets.

antonbabenko:
There have been multiple issues (feature requests) like this. The resolution was (and still is) that we don't want to have a new type of service-related subnets in this module before we upgrade the module to use `for_each` (among a few other things). When exactly will it happen, we don't know yet.

Closing this issue for now.

PR #878, opened 2023-01-06
feat: Make NAT Gateway Routes for Private Route Tables Optional

cmccabe-ws:
## Description
Added a new variable which will allow for NAT gateway routes on private route tables to be skipped.

## Motivation and Context

I have a scenario in my environment where I need to override the 0.0.0.0/0 route for the private route tables.

For example:

I want to deploy the AWS Network Firewall in my VPC that I have built using this module. I need to update my private subnets 0.0.0.0/0 route to use the network firewall endpoint rather than the NAT Gateway endpoint.

I understand that the new version (V4) of terraform-aws-vpc will have network functionality baked in, however in the interim it would be nice to be able to override this route creation.

## Breaking Changes
By default this will be a no-op, as the default is set to true, which the count expects in order to create the routes.

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1049,7 +1049,7 @@ resource "aws_nat_gateway" "this" {
 }
 
 resource "aws_route" "private_nat_gateway" {
-  count = local.create_vpc && var.enable_nat_gateway ? local.nat_gateway_count : 0
+  count = local.create_vpc && var.enable_nat_gateway && var.create_private_nat_gateway_route ? local.nat_gateway_count : 0
 
   route_table_id         = element(aws_route_table.private[*].id, count.index)
   destination_cidr_block = var.nat_gateway_destination_cidr_block
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ If `one_nat_gateway_per_az = true` and `single_nat_gateway = false`, then the mo
 
 By default, if NAT Gateways are enabled, private subnets will be configured with routes for Internet traffic that point at the NAT Gateways configured by use of the above options.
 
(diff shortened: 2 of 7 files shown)


cmccabe-ws:
👋 Hey @antonbabenko or @bryantbiggs could I get a review on this change?

main.tf
```diff
@@ -1051,7 +1051,7 @@ resource "aws_nat_gateway" "this" {
 }
 
 resource "aws_route" "private_nat_gateway" {
-  count = local.create_vpc && var.enable_nat_gateway ? local.nat_gateway_count : 0
+  count = local.create_vpc && var.enable_nat_gateway && var.create_private_nat_gateway_route ? local.nat_gateway_count : 0
```
bryantbiggs, line 1054: I don't follow the logic here - you want to create a NAT GW but you don't want to route it?

cmccabe-ws: @bryantbiggs I can understand why this is hard to follow - 

I do want to route to it, but when running AWS Network Firewall, traffic should generally flow:

private host -> network firewall -> NAT -> IGW

So in order to support that, the 0.0.0.0/0 route on the private route table must actually route to the VPC endpoint ID of the network firewall, not the NAT gateway.

Since this module does not currently support network firewall, the easiest way to accomplish that is to override the creation of the 0.0.0.0/0 NAT route in the module, and create the 0.0.0.0 NFW route outside of it - this way there are no conflicts.

```hcl
module example {
  source = "some-source"
   . . . 
  enable_nat_gateway = true
  create_private_nat_gateway_route = false
}

resource "aws_route" "network_firewall_route" {
  for_each               = toset(concat(module.vpc.private_route_table_ids))
  route_table_id         = each.value
  destination_cidr_block = "0.0.0.0/0"
  vpc_endpoint_id        = module.network_firewall.network_firewall_endpoint_id[0]
}

```

The network firewall's route table would actually have the 0.0.0.0/0 route to the NAT Gateway.

cmccabe-ws: @bryantbiggs any update here?

cmccabe-ws:
Ended up forking repository to make the changes, though I do think lack of support network firewall in the module (or providing a mechanism to modify routes to support it as per this PR) does not align with security best practices.

Issue #883, opened 2023-01-16
Manage Internet Gateway attachment explicitly to avoid dependency issue


onitake:
## Is your request related to a new offering from AWS?
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - Yes ✅: 4.3.0

## Is your request related to a problem? Please describe.

This is a follow-up to #283 with a proposed solution.

#283 was closed, but there was no solution provided for dependencies created by this module, such as the Internet Gateway. The only suggestion was to delete resources manually, but this can be a problem in scenarios where infrastructure is provisioned automatically. The module should at least be able to handle dependencies on resources that are local to the module.

## Describe the solution you'd like.

With the new [internet_gateway_attachment](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/internet_gateway_attachment) resource implemented in https://github.com/hashicorp/terraform-provider-aws/pull/16386 , it should now be possible to resolve this dependency issue, at least for cases where no external dependencies on the VPC are involved.

I suggest replacing the [IGW creation + attachment code](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/90ef90a311c547dd05054a81d2de592885ba9996/main.tf#L127-L137) with the following:

```terraform
resource "aws_internet_gateway" "this" {
  count = local.create_vpc && var.create_igw && length(var.public_subnets) > 0 ? 1 : 0

  tags = merge(
    { "Name" = var.name },
    var.tags,
    var.igw_tags,
  )
}

resource "aws_internet_gateway_attachment" "this" {
  count = local.create_vpc && var.create_igw && length(var.public_subnets) > 0 ? 1 : 0
  internet_gateway_id = aws_internet_gateway.this[0].id
  vpc_id = local.vpc_id
}
```

## Describe alternatives you've considered.

* Creating the Internet Gateway outside the VPC module (redundant)
* Manual detachment/deletion (can be difficult in certain scenarios)

## Additional context

Upstream issue for the IGW/VPC dependency handling bug: https://github.com/hashicorp/terraform-provider-aws/issues/16151

onitake:
I implemented this in #896 because it's still a major issue for me, but I still need to test it on one of the examples (hence the WIP status).

onitake:
[...]

PR #896, opened 2023-02-20
fix: WIP: Separate IGW creation from VPC attachment

onitake:
## Description

This PR separates creation of the internet gateway (IGW) from attachment to the VPC.

By doing this, an implicit dependency is avoided, allowing `terraform destroy` on a VPC+IGW created by this module.

## Motivation and Context

Fixes: #883 

## Breaking Changes

The new resource `aws_internet_gateway_attachment` requires version 4.3.0 of the Terraform provider for AWS.

## How Has This Been Tested?


onitake:
As written in #883 , I couldn't reproduce the issue with the simple-vpc example and a custom test case with the current version of the AWS provider. However, I'm still experiencing the dependency problem in a production environment.
Further testing is needed.

PR #899, opened 2023-02-27
improvement: allow custom network ACL for specific subnets

### LINKED ISSUE #898: Allow network ACL management exclusion for given subnets
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - Yes ✅: AWS provider v3+ (I think)

We would like to test an AZ failure for Business Continuity Plan (BCP) or Chaos engineering.

To simulate an AZ failure, a solution is to flip the network ACL for all the subnets in an AZ and set `DENY` traffic to all.

We could introduce an input variable to exclude network ACL management for some given subnets.

```hcl
module "vpc" {
  source = "terraform-aws-modules/vpc/aws"

  private_dedicated_network_acl = true
  private_acl_excluded_subnets  = ["subnet-123456789"]

  # ...
}

resource "aws_network_acl" "block_all" {
  vpc_id     = module.vpc.vpc.id
  subnet_ids = ["subnet-123456789"]

  ingress {
    rule_no    = 1
    protocol   = "all"
    from_port  = 0
    to_port    = 0
    cidr_block = "0.0.0.0/0"
    action     = "deny"
  }

  ingress {
    rule_no         = 2
    protocol        = "all"
    from_port       = 0
    to_port         = 0
    ipv6_cidr_block = "::/0"
    action          = "deny"
  }

  egress {
    rule_no    = 1
    protocol   = "all"
    from_port  = 0
    to_port    = 0
    cidr_block = "0.0.0.0/0"
    action     = "deny"
  }

  egress {
    rule_no         = 2
    protocol        = "all"
    from_port       = 0
    to_port         = 0
    ipv6_cidr_block = "::/0"
    action          = "deny"
  }
}
```

I don't see any other alternative.

## Additional context

Current:
Page 1

BCP test with one AZ failure simulation:
BPC

samuel-phan:
fix #898

## How Has This Been Tested?


main.tf
```diff
@@ -650,7 +650,7 @@ resource "aws_network_acl" "public" {
   count = local.create_vpc && var.public_dedicated_network_acl && length(var.public_subnets) > 0 ? 1 : 0
 
   vpc_id     = local.vpc_id
-  subnet_ids = aws_subnet.public[*].id
+  subnet_ids = [for sid in aws_subnet.public[*].id : sid if !contains(var.public_network_acl_excluded_subnets, sid)]
```
bryantbiggs, line 653: you cannot pass subnet IDs that you do not yet know their value - this is incorrect

bryantbiggs:
this is an incorrect implementation - we can continue to discuss in the issue linked before any changes are proposed

main.tf
samuel-phan: In general, yes, but here, it's for a special use case where one has executed the VPC module first to create the resources, and afterwards simulate an AZ outage for a specific subnet.

The best way to target without any ambiguity a subnet is through the subnet ID.

We could offer as an alternative filter based on tags also, but the filter by subnet ID should still be an option, I think.

bryantbiggs: >  but here, it's for a special use case

And this is why I don't believe this type of change belongs in the module

samuel-phan: How could we test an AZ outage then? Isn't it a use case that the VPC module would/should support? Or is there any other way to test AZ outage?

I'm all ears, if there's an alternative.

PR #908, opened 2023-03-10
chore: Updated example to enable TLS Inbound for Session Manager

kjenney:
Enable TLS inbound for Session Manager

## Description
I use Session Manager quite extensively as it avoids the need for managing SSH Key pairs. The complete example had all of the endpoints but was missing the security group definitions for a couple, so I added them here.

## Motivation and Context
Enable testers to spin up EC2 instances without the need for SSH Key Pairs

## Breaking Changes
None

## How Has This Been Tested?
I've validated that this works in 2 separate AWS accounts.


antonbabenko:
This PR is included in version 4.0.0 :tada:

PR #911, opened 2023-03-15
Fix cidr_block set to null if use_ipam_pool set to true

osaluden:
## Description
*cidr_block* parameter is set to `null` when *use_ipam_pool* varaiable is set to true. Thus, only *ipv4_netmask_length* is available when you want to use IPAM. However, it is possible to chose the CIDR rather than the netmask length as the [example](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/examples/ipam-vpc/main.tf) shows at line 44. 

## Motivation and Context
Want to be able to use IPAM with CIDR selection, not only with netmask length selection.

## Breaking Changes
No breaking changes.

## How Has This Been Tested?
- [x] Test with *cidr* provisioned as [example](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/examples/ipam-vpc/main.tf) shows at line 44
- [x] Test with *ipv4_netmask_length* provisioned as [example](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/examples/ipam-vpc/main.tf) shows at line 24
- [x] Test with *use_ipam_pool* set to *false*

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -20,7 +20,7 @@ locals {
 resource "aws_vpc" "this" {
   count = local.create_vpc ? 1 : 0
 
-  cidr_block          = var.use_ipam_pool ? null : var.cidr
+  cidr_block          = var.use_ipam_pool && var.ipv4_netmask_length != null ? null : var.cidr
   ipv4_ipam_pool_id   = var.ipv4_ipam_pool_id
   ipv4_netmask_length = var.ipv4_netmask_length
 


osaluden:
Hi there

After several tests an deployment, I discovered that if the parameter *use_ipam_pool* is set to false, it is still possible to use the *ipv4_ipam_pool_id* parameter to register the VPC CIDR into the desired IPAM pool. This way, the *cidr* parameter can be used as well as *ipv4_netmask_length* if you want to ask IPAM to provided you with a CIDR range with the desired netmask length.  

Let's close the pull request.

PR #979, opened 2023-09-07
docs: Add missing 'Allow' statement in example/complete.tf VPC endpoint policy

andormarkus:
## Description
Add missing 'Allow' statement in example/complete.tf VPC endpoint policy. This omission could lead to hard-to-debug errors.

## Motivation and Context
We encountered a challenging error due to the sample policy missing an 'Allow' statement. I've updated the sample to include this, helping others avoid the same issue.

## Breaking Changes
None

## How Has This Been Tested?

### DIFF
--- a/examples/complete/main.tf
+++ b/examples/complete/main.tf
@@ -28,20 +28,36 @@ module "vpc" {
   name = local.name
   cidr = local.vpc_cidr
 
-  azs                 = local.azs
-  private_subnets     = [for k, v in local.azs : cidrsubnet(local.vpc_cidr, 8, k)]
-  public_subnets      = [for k, v in local.azs : cidrsubnet(local.vpc_cidr, 8, k + 4)]
-  database_subnets    = [for k, v in local.azs : cidrsubnet(local.vpc_cidr, 8, k + 8)]
-  elasticache_subnets = [for k, v in local.azs : cidrsubnet(local.vpc_cidr, 8, k + 12)]
-  redshift_subnets    = [for k, v in local.azs : cidrsubnet(local.vpc_cidr, 8, k + 16)]
-  intra_subnets       = [for k, v in local.azs : cidrsubnet(local.vpc_cidr, 8, k + 20)]
+  azs = local.azs
+  private_subnets = [
+    for k, v in local.azs : cidrsubnet(local.vpc_cidr, 8, k)
+  ]
+  public_subnets = [
+    for k, v in local.azs : cidrsubnet(local.vpc_cidr, 8, k + 4)
+  ]
+  database_subnets = [
+    for k, v in local.azs : cidrsubnet(local.vpc_cidr, 8, k + 8)
+  ]
+  elasticache_subnets = [
+    for k, v in local.azs : cidrsubnet(local.vpc_cidr, 8, k + 12)
+  ]
+  redshift_subnets = [
+    for k, v in local.azs : cidrsubnet(local.vpc_cidr, 8, k + 16)
+  ]
+  intra_subnets = [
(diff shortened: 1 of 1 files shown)


antonbabenko (commented):
https://docs.aws.amazon.com/vpc/latest/privatelink/concepts.html 
> A VPC endpoint policy is an IAM resource policy that you attach to a VPC endpoint. It determines which principals can use the VPC endpoint to access the endpoint service. The default VPC endpoint policy allows all actions by all principals on all resources over the VPC endpoint.

So, I don't think this `Allow` statement is necessary.

bryantbiggs:
also - this is just an example of providing a policy, not necessarily telling users what that policy should be. I don't think any changes are warranted here

andormarkus:
> https://docs.aws.amazon.com/vpc/latest/privatelink/concepts.html
> 
> > A VPC endpoint policy is an IAM resource policy that you attach to a VPC endpoint. It determines which principals can use the VPC endpoint to access the endpoint service. The default VPC endpoint policy allows all actions by all principals on all resources over the VPC endpoint.
> 
> So, I don't think this `Allow` statement is necessary.


The default policy is set to 'Full access'. However, if you implement a custom policy as demonstrated here, it will override the default. Consequently, attempting to use the endpoint with this custom policy will result in the error described below.
```shell
sh-X.X$ aws sns list-topics

An error occurred (AuthorizationError) when calling the ListTopics operation: 
User: arn:aws:sts::[account-id]:assumed-role/[role-name]/[instance-id] is not authorized to perform: 
SNS:ListTopics on resource: arn:aws:sns:[region]:[account-id]:* because no VPC endpoint policy allows 
the SNS:ListTopics action
```
Used VPC endpoint policy
```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Deny",
      "Action": "*",
      "Resource": "*",
      "Principal": "*",
      "Condition": {
        "StringNotEquals": {
          "aws:SourceVpc": [ vpc-id ]
        }
      }
    }
  ]
}
```

Once I have added the allow policy it worked
```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Deny",
      "Action": "*",
      "Resource": "*",
      "Principal": "*",
      "Condition": {
        "StringNotEquals": {
          "aws:SourceVpc": [vpc-id]
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": "*",
      "Resource": "*",
      "Principal": "*",
      "Condition": {
        "StringEquals": {
          "aws:SourceVpc": [vpc-id]
        }
      }
    }
  ]
}

```

```shell
sh-5.2$ aws sns list-topics
{
    "Topics": [
        {
            "TopicArn": "arn:aws:sns:[region]:[account-id]:topic_1"
        }
    ]
}
```

andormarkus:
> also - this is just an example of providing a policy, not necessarily telling users what that policy should be. I don't think any changes are warranted here

While I understand that this is just an example, we need to acknowledge a critical nuance: deploying this example via Terraform will technically complete without errors, but users will be met with challenges when trying to use the VPC endpoint due to the incorrect policy. 

Furthermore, while the AWS CLI and API might return easily understandable error codes like I showed in my previous comment, other services do not exhibit the same transparency. For instance, on the AWS console, the SSM session manager will simply fail to start a session, providing a vague message such as 'Can't connect to your instance.' Similarly, MWAA will not launch; instead, after a 2-hour deploying, it will fail with an 'incorrect configuration' error.

This examples serve as a reference for users, and any ambiguity or oversight could significantly impact their experience. It's not just about successful deployment; the usability of the provisioned resources is paramount. Moreover, ensuring clarity might be beneficial for users less familiar with AWS policies.

andormarkus:
Remove stale

PR #985, opened 2023-09-24
fix: Ignore additional tags added for VPC and subnets outside the module.

gohmc:
## Description

Currently, additional tags added to VPC and subnet resources outside the module will be _remove_ when re-apply, even there's no change to the network topology. To prevent tags from removal unexpectedly, this PR added lifecycle `ignore_changes` to `tags` for resource `aws_vpc` and `aws_subnet`.

- Applied any of the [example](https://github.com/terraform-aws-modules/terraform-aws-vpc/tree/master/examples/simple)
- Manually add tags to the VPC and subnets created by this module. This simulate an external process (eg. INFOSEC) adding additional tags.
- Re-apply the same example and will see terraform plan to remove additional tags added outside the module.

## Motivation and Context

Additional tags can be added post applied.

## Breaking Changes

No.

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -48,6 +48,10 @@ resource "aws_vpc" "this" {
     var.tags,
     var.vpc_tags,
   )
+
+  lifecycle {
+    ignore_changes = [tags]
+  }
 }
 
 resource "aws_vpc_ipv4_cidr_block_association" "this" {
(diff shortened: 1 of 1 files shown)


antonbabenko:
This will not make it into the module because we want users to be able to manage tags across resources created by this module.

Closes #986

PR #991, opened 2023-10-05
fix: Update intra subnets route tables to be one per AZ

megashby:
## Description
Create one route table and associated route table association per subnet for intra subnets

## Motivation and Context
We need to deploy lambdas in a VPC that has both private subnet and intra subnet tiers. Our use case is some lambdas will need to be able to connect to the internet via NATGWs, but some lambdas won't need any outbound internet access. For lambdas that will be deployed in the intra subnets, they will be in multiple AZs and the current pattern of only having one route table shared amongst all subnets is not compatible with patterns like having VPC endpoints / private link endpoints deployed in all subnets.

## Breaking Changes
outputs for `intra_subnets_route_table_ids` and` intra_subnets_route_table_association_ids` will now potentially have more than 1 output.

## How Has This Been Tested?
- [ x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [x ] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [ x] I have executed `pre-commit run -a` on my pull request

Running complete and separate-route-tables examples show that 3 route tables are now being created.


lslaslo:
👍

PR #994, opened 2023-10-21
fix: Remove references to the create_vpc variable from unrelated reso…

Yatin03:
…urces such as subnets, route tables, and NAT gateways. If 'create_vpc' variable value is false then you have to provide vpc_id.

## Description
If i set create_vpc variable value to false then it doesn't create other resources like: Subnet, Route Table, NAT Gateway.

## Motivation and Context
Currently, Setting the create_vpc variable value to "false" ensures that the creation of additional resources such as subnets, route tables, and NAT gateways is avoided. This change is required to provide more flexibility and control over resource provisioning. It allows for a more customized deployment process, particularly in scenarios where you may already have existing resources or you want to manage these specific resources separately. By making this change, you can prevent the unnecessary creation of these resources when they are not needed, which can save resources, reduce complexity, and better align with your specific requirements.

## Breaking Changes
No, It should not break existing cluster because i have changed creation conditions

## How Has This Been Tested?
- [X ] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [ X] I have tested and validated these changes using one or more of the provided `examples/*` projects
image

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -10,16 +10,18 @@ locals {
   max_subnet_length = max(
     local.len_private_subnets,
     local.len_public_subnets,
+    local.len_intra_subnets,
     local.len_elasticache_subnets,
     local.len_database_subnets,
     local.len_redshift_subnets,
   )
 
   # Use `local.vpc_id` to give a hint to Terraform that subnets should be deleted before secondary CIDR blocks can be free!
-  vpc_id = try(aws_vpc_ipv4_cidr_block_association.this[0].vpc_id, aws_vpc.this[0].id, "")
+  vpc_id = try(aws_vpc_ipv4_cidr_block_association.this[0].vpc_id, aws_vpc.this[0].id, var.vpc_id)
 
-  create_vpc = var.create_vpc && var.putin_khuylo
+  create_vpc = var.create_vpc
 }
+
 
 ################################################################################
 # VPC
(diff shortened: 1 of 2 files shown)


bryantbiggs:
> create_vpc variable value to false then it doesn't create other resources like: Subnet, Route Table, NAT Gateway.

this is the `vpc` module, its intent is to create the VPC and all of the supporting networking components that go along with the VPC.

Yatin03:
> > create_vpc variable value to false then it doesn't create other resources like: Subnet, Route Table, NAT Gateway.
> 
> this is the `vpc` module, its intent is to create the VPC and all of the supporting networking components that go along with the VPC.

Hi Bryant,

I agree that this is a VPC module. However, I've noticed that there's a create_vpc variable within this module. If I set it to "false," it creates a random string resource, which seems unnecessary.

I recently encountered a use case where I already had an existing VPC, and my intention was to create additional resources such as subnets and route tables within that VPC but that is not possible using this module. It's possible that other team members have faced similar situations in the past.

Considering this, I felt it was important to address this issue and make this change. This modification is necessary to enhance the flexibility and control over resource provisioning, allowing us to work with existing VPCs more efficiently.

antonbabenko:
@Yatin03 Consider importing your existing VPC to be managed by the module. This PR introduces a feature that will be very hard to maintain in the long run, so we can't have it.

Yatin03:
@antonbabenko, This modification won't disrupt the importing feature. 'create_vpc' variable was already present, but it was not functioning as expected. Therefore, I have made some adjustments. Could you please review it?

bryantbiggs:
It is functioning as expected - if you set create_vpc to false, the VPC and nearly all resources are not created. That's how it's designed

PR #1016, opened 2023-12-07
feat: Add multiple intra route tables

jbrooks-sms:
## Description
This adds multiple intra route tables per the number of intra subnets deployed.

## Motivation and Context
This resolves an issue where multiple intra route tables are required such as when using them in a multiple availability zone inspection vpc.  Each intra subnet needs to forward to the particular network firewall endpoint which is in the local availability zone

## Breaking Changes
There are no breaking changes as this only creates the same number of route tables as intra subnets.  

## How Has This Been Tested?
I have tested by running the seperate-route-tables example in this branch and outputs return everything as expected to include the multiple route table ids.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -817,12 +817,18 @@ resource "aws_subnet" "intra" {
 }
 
 resource "aws_route_table" "intra" {
-  count = local.create_intra_subnets ? 1 : 0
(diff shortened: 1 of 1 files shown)


aserakhi:
Very intrested with this feature, we are facing the same issues to manage routing NFW endpoints per AZ!

PR #1021, opened 2023-12-12
chore: Fix misconfigured IAM policies in examples/complete

enm10k:
## Description

Fixed misconfigured IAM policies in examples/complete

## Motivation and Context

I created a VPC Endpoints based on examples/complete, but it did not work due to misconfigured IAM policies.

## How Has This Been Tested?

### DIFF
--- a/examples/complete/main.tf
+++ b/examples/complete/main.tf
@@ -161,7 +161,7 @@ module "vpc_endpoints_nocreate" {
 
 data "aws_iam_policy_document" "dynamodb_endpoint_policy" {
   statement {
-    effect    = "Deny"
+    effect    = "Allow"
     actions   = ["dynamodb:*"]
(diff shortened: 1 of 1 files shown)


bryantbiggs:
they are not mis-configured because they are merely examples - users will need to adjust the policies based on their business requirements

said differently - we have examples that create S3 buckets but the buckets are not encrypted. Is this mis-configured? I would argue no because there is no context other than they are examples for demonstration and testing purposes; users will need to configure the appropriate settings for their individual requirements

enm10k:
Thanks for your explanation.
I understand what you are saying.

At first glance, the current IAM policies seem to allow operations from the VPC, but it actually operates to deny all actions.
I thought it was confusing, but as you say, it might not be a problem as examples.

Issue #1020, opened 2023-12-11
support for multipe public route tables


pedrojflores:
## Is your request related to a new offering from AWS?
Yes.  This is to better support a vpc hosting an AWS Firewall infrastructure where you need to route traffic from public subnets to an unique firewall interface on a specific az.

## Is your request related to a problem? Please describe.
Yes.  Due to this limitation I can't use this module to build my AWS Firewall Egress vpcs.

## Describe the solution you'd like.
I would like an option that when set a unique route table gets created per public subnet (per az).  Same support the module has for private subnets.

nirmesika:
add support for adding routes to public and private subnets.

piartz:
@pedrojflores I have created a pull request, please let me know if this is what you were looking for, and we can wait for a review. 
https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/1025

Feliz año 2024!

piartz:
Pumping up this issue as it is not stale anymore

KevinSnyderCodes:
I've also created a pull request that expands this to both public and intra subnets: https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/1051

I punted on the other subnet types as their logic was a bit more complex, but it should be possible to support multiple route tables for all subnet types.

KevinSnyderCodes:
Not stale

antonbabenko:
This issue has been resolved in version 5.8.0 :tada:

PR #1025, opened 2024-01-04
feat: Support for multiple public route tables

piartz:
## Description
Support for multiple route tables when creating public subnets

## Motivation and Context
Guarantee that the used route table for public subnets on creation is different when a boolean variable is set to true.
Motivation from issue https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/1020 
## Breaking Changes
No breaking changes, as the variable is optional and defaulted to 'false'. 

## How Has This Been Tested?

Used the examples/simple/* directory as a start. Changed the `private_subnets` array for `public_subnets` and verified that the behavior described by the requester is happening. Included the variable `create_multiple_public_route_tables` and set to true. Verified in multiple runs of apply and destroy that the created route table is different for each new public subnet. 

Changes to examples/simple/ have not been reflected on the push. I also did not want to copy-paste the contents of 'simple' for this use case, for a more clean, less complex state of the examples folder. Please let me know if you prefer the example to be included.


piartz:
upping the PR so it is not stale anymore

PR #1026, opened 2024-01-04
feat: Add validation to limit secondary_cidr_blocks to max 4 entries

mck-dpadmanabhan:
## Description
Add validation to limit secondary_cidr_blocks to max 4 entries

## Motivation and Context
- Introduce a validation block within the variable 'secondary_cidr_blocks'
- Ensure the list of CIDR blocks does not exceed 4 entries
- Provide a clear error message for configurations exceeding the limit

## How Has This Been Tested?


antonbabenko:
We don't need such a validation block because we expect users to specify what is necessary.

PR #1056, opened 2024-03-22
feat: Allow setting vpc endpoints as an input for each endpoint

### LINKED ISSUE #1054: VPC endpoints are being replaced sporadically due to `service_name` being unknown at plan time
## Description

vpc endpoints flapping sporadically due to https://github.com/hashicorp/terraform-provider-aws/issues/25568. 

Commented on problem here:
https://github.com/hashicorp/terraform-provider-aws/issues/25568#issuecomment-1998004128

> This is happening for us sporadically. It doesn't happen all the time. Sometimes we apply and suddenly all of our VPC endpoints need to be replaced. We aren't changing anything going into the VPC module or relating to the endpoints. Sidenote - we are deploying in govcloud.
>
> We're using this module:
> 
> https://github.com/defenseunicorns/terraform-aws-vpc/blob/main/main.tf#L130-L256 which feeds into this ->
> https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/modules/vpc-endpoints/main.tf#L11-L21
> 
> truncated output, but it does this for all of our endpoints that get fed from that `aws_vpc_endpoint_service` data that feed into the `aws_vpc_endpoint` resource
> 
> ```hcl
>          ~ requester_managed     = false -> (known after apply)                                                                                                                  
>            ~ route_table_ids       = [] -> (known after apply)                                                                                                                     
>            ~ service_name          = "com.amazonaws.us-gov-west-1.ssm" # forces replacement -> (known after apply) # forces replacement                                            
>            ~ state                 = "available" -> (known after apply)    
> ```

## Versions

- Module version [Required]:
v5.5.3

- Terraform version:
Terraform v1.5.7
on darwin_arm64

- Provider version(s):
aws 2.4.1

[defenseunicorns/terraform-aws-vpc@main/main.tf#L130-L256](https://github.com/defenseunicorns/terraform-aws-vpc/blob/main/main.tf#L130-L256) which feeds into this ->
[terraform-aws-modules/terraform-aws-vpc@master/modules/vpc-endpoints/main.tf#L11-L21](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/modules/vpc-endpoints/main.tf#L11-L21)

1.  Deploy vpc + endpoints using https://github.com/defenseunicorns/terraform-aws-vpc which uses this module
2. Deploy again some time later. TF will detect that the service names are unknown and replace them for no reason.

## Expected behavior

They should not be replaced if nothing is changing.

## Actual behavior

The VPC endpoints are forced to be replace due to the use of the `aws_vpc_endpoint_service` data source, sometimes. This is due to them being known after apply.

### Terminal Output Screenshot(s)

basically this, for all VPC endpoints, sometimes.

image

### possible solution

Ideally this module should allow the use of feeding in a `service_name` and doing a `lookup()` on that so that we can build it outside when passing in the map of `endpoints`, but defaulting to `data.aws_vpc_endpoint_service.this[each.key].service_name` if none provided.

zack-is-cool:
## Description

Adds feature to allow overriding for each `aws_vpc_endpoint` `service_name` property, but defaulting to data if not defined.

Additionally this allows users to use specify fips endpoints if needed.

## Motivation and Context

fixes #1054 

## Breaking Changes

N/A

## How Has This Been Tested?

Have a PR here for an example where this was tested using changes in my `.terraform` directory which I replicated in this fork.
https://github.com/defenseunicorns/terraform-aws-vpc/pull/113

### DIFF
--- a/modules/vpc-endpoints/main.tf
+++ b/modules/vpc-endpoints/main.tf
@@ -24,7 +24,7 @@ resource "aws_vpc_endpoint" "this" {
   for_each = local.endpoints
 
   vpc_id            = var.vpc_id
-  service_name      = data.aws_vpc_endpoint_service.this[each.key].service_name
+  service_name      = try(each.value.service_endpoint_url, data.aws_vpc_endpoint_service.this[each.key].service_name)
   vpc_endpoint_type = try(each.value.service_type, "Interface")
   auto_accept       = try(each.value.auto_accept, null)
 
--- a/modules/vpc-endpoints/main.tf
+++ b/modules/vpc-endpoints/main.tf
@@ -24,7 +24,7 @@ resource "aws_vpc_endpoint" "this" {
   for_each = local.endpoints
 
   vpc_id            = var.vpc_id
-  service_name      = try(each.value.service_endpoint_url, data.aws_vpc_endpoint_service.this[each.key].service_name)
+  service_name      = try(each.value.service_endpoint, data.aws_vpc_endpoint_service.this[each.key].service_name)
   vpc_endpoint_type = try(each.value.service_type, "Interface")
   auto_accept       = try(each.value.auto_accept, null)
 


modules/vpc-endpoints/main.tf
```diff
@@ -24,7 +24,7 @@ resource "aws_vpc_endpoint" "this" {
   for_each = local.endpoints
 
   vpc_id            = var.vpc_id
-  service_name      = data.aws_vpc_endpoint_service.this[each.key].service_name
+  service_name      = try(each.value.service_endpoint_url, data.aws_vpc_endpoint_service.this[each.key].service_name)
```
antonbabenko, line 27: How `service_endpoint_url` should look for the FIPS environment?

zack-is-cool: Sorry, I wasn't clear on this, I meant more so in the context of since you can override with inputs, users can use fips endpoints

see:
https://github.com/defenseunicorns/terraform-aws-vpc/pull/113/files#diff-dc46acf24afd63ef8c556b77c126ccc6e578bc87e3aa09a931f33d9bf2532fbbR227

You'd probably able to do it with this module if we added another `data "aws_vpc_endpoint_service"` with different filters, but It doesn't seem like a super common enough request.

antonbabenko: I see. I think it should be named `service_endpoint` instead of `service_endpoint_url` to represent the actual type of the expected value. Could you change it and I will merge it? :)

antonbabenko (changes requested):

modules/vpc-endpoints/main.tf
zack-is-cool: Sure, no problem.

antonbabenko (approved):

antonbabenko:
This PR is included in version 5.7.0 :tada:

Issue #1091, opened 2024-06-13
VPC Endpoints do not support `ip_address_type` property


adam-sm:
## Description

Setting the [ip_address_type](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_endpoint#ip_address_type)  property on an endpoint does not get propagated to the underlying `aws_vpc_endpoint` resource.

This means the routing tables only get propagated with the S3/DynamoDB IPv4 prefix list routes and clients using IPv6 (AWS CLI) can not route request to S3 and hang if no external egress  

- [x ] ✋ I have searched the open/closed issues and my issue is not listed.

tokarev-artem:
Hi @adam-sm , can you check if #1096 works for you?

tokarev-artem:
@adam-sm up

bryantbiggs:
resolved in #1096

PR #1096, opened 2024-06-17
feat: Add support for `ip_address_type` for VPC endpoint

tokarev-artem:
## Description
Added parameter ip_address_type to allow ipv4. ipv6 and dualstack ip_address_type support

## Motivation and Context
To solve issue #1091 

## Breaking Changes
None

## How Has This Been Tested?
- [*] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [*] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [*] I have executed `pre-commit run -a` on my pull request

### DIFF
--- a/examples/ipv6-dualstack/main.tf
+++ b/examples/ipv6-dualstack/main.tf
@@ -47,3 +47,36 @@ module "vpc" {
 
   tags = local.tags
 }
+
+module "vpc_endpoints" {
+  source = "../../modules/vpc-endpoints"
+
+  vpc_id = module.vpc.vpc_id
+
+  create_security_group      = true
+  security_group_name_prefix = "${local.name}-vpc-endpoints-"
+  security_group_description = "VPC endpoint security group"
+  security_group_rules = {
+    ingress_https = {
+      description = "HTTPS from VPC"
+      cidr_blocks = [module.vpc.vpc_cidr_block]
+    }
+  }
+
+  endpoints = {
+    athena = {
+      service         = "athena"
+      ip_address_type = "dualstack"
+      tags            = { Name = "athena-vpc-endpoint" }
+    },
+    lakeformation = {
+      service         = "lakeformation"
+      ip_address_type = "ipv6"
+      tags            = { Name = "lakeformation-vpc-endpoint" }
+    },
+    ecs = {
+      service = "ecs"
+      tags    = { Name = "ecs-vpc-endpoint" }
+    },
+  }
+}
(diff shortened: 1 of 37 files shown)


examples/ipv6-dualstack/main.tf
```diff
@@ -47,3 +47,36 @@ module "vpc" {
 
   tags = local.tags
 }
+
+module "vpc_endpoints" {
```
bryantbiggs, line 51: we don't need a new example - we can use an existing example

bryantbiggs (changes requested):

examples/ipv6-dualstack/main.tf
tokarev-artem: hi @bryantbiggs , I have a question, 
as an example - ipv6 VPC endpoints can be created without subnets attached. But as a good example - probably better to add a fully working example.  If the second option is preferable - we need to add ipv6 subnets to the VPC setup. Something like that: 
```
    athena = {
      service                 = "athena"
      ip_address_type = "dualstack"
      subnet_ids           = [ipv6_subnet1, ipv6_subnet2. ipv6_subnet3]
      tags                       = { Name = "athena-vpc-endpoint" }
    },
```
What do you think about this ?

bryantbiggs: apologies, I don't understand the question - what are you asking?

tokarev-artem: sorry, let me rephrase with details

I can add this minimal example, which will create the endpoint but it will not work because subnets are not defined
```
lakeformation = {
  service         = "lakeformation"
  ip_address_type = "ipv6"
  tags            = { Name = "lakeformation-vpc-endpoint" }
}
```
or add an example like this, which is not correct, because `module.vpc.private_subnets` returns all subnets (ipv4 and ipv6), but not only ipv6
```
lakeformation = {
  service         = "lakeformation"
  ip_address_type = "ipv6"
  subnet_ids      = module.vpc.private_subnets
  tags            = { Name = "lakeformation-vpc-endpoint" }
},
```
so the question is - what do you think, is there any good workaround ?

bryantbiggs: got it - we can forgo updating the example for this, its just adding one argument thats a direct pass through, not much to test/validate/demonstrate

tokarev-artem: Removed the example

bryantbiggs (approved):

antonbabenko:
This PR is included in version 5.13.0 :tada:

PR #1109, opened 2024-08-02
chore: Support secondary eips 

### LINKED ISSUE #1108: Support secondary ip addresses for nat gateway
The functionality is there, I don't know when it was introduced, but I presume for at least a year.

We are being rate limited by github based on IP address from one of our aws accounts, in order to work around this problem we added secondary ip addresses to our natgateways.  We have no way of configuring the vpc module to reflect this desired configuration.

Expose configuration for secondary ip addresses to allow more than 1 ip address to be configured per nat-gateway

configuring the nat-gateway outside of this module.

AlexisColes:
resolves #1108


main.tf
```diff
@@ -1074,6 +1075,22 @@ resource "aws_eip" "nat" {
   depends_on = [aws_internet_gateway.this]
 }
 
+resource "aws_eip" "secondary" {
+  for_each = toset(flatten([for nat in aws_eip.nat : [for suffix in local.seips_suffixs : "${nat.tags.Name}-${suffix}"]]))
```
bryantbiggs, line 1079: we cannot have computed values as keys in maps

bryantbiggs (changes requested):

main.tf
AlexisColes: Hey, could you explain why please and if you could recommend an alternative approach?

I did consider doing this with count like the other arrays were doing, however if you changed the number of AZs or number of ips per az it would start trying to move the ips from 1 natgateway to the other which then wants to destroy and re-create the natgateway which isn't ideal.  Using keys seem to make the solution much more solid.

Or is it the use of the `nat.tags.Name` that is the issue.  This does seem like it could be flaky :thinking:  I guess we could use a range based on 
`local.create_vpc && var.enable_nat_gateway && !var.reuse_nat_ips ? local.nat_gateway_count : 0`
As the first part of the composite key, this would be much more solid thinking about it :rocket:

flaviomoringa: The above errors happened running the code in a previously created VPC, with 3 NATGW (1 per az), and changing the secondary IP's from 0 to 2.

flaviomoringa: I'm testing the following:

Adding to locals:
  seips_names = flatten([for nat_index in range(0, local.nat_gateway_count) : [for suffix in local.seips_suffixs : "${aws_eip.nat[nat_index].tags.Name}-${suffix}"]])

and replacing the for_each with:
for_each = toset(local.seips_names)

Would that help?

flaviomoringa: guess not... same errors "Error: Provider produced inconsistent final plan" when going from 0 secondary IP's to 2 :-(

velkovb: @flaviomoringa I have done some testing and it seems a provider issue indeed. The only way to add secondary IPs with terraform is on NAT Gateway creation. Any further change to that results in a re-creation of the NAT Gateway. However, if you have created it without secondary IPs you get the inconsistent plan error. I believe that has to do with how it associates the private IPs to the NAT Gateway interfaces but it doesn't allow manual association.

I tested with this very basic example:
```hcl
locals {
  nat_gateway_count                    = 2
  public_subnets                       = ["subnet-xxxxxx", "subnet-xxxxxx"]
  number_of_secondary_eips_per_gateway = 0
}

resource "aws_eip" "nat" {
  count = local.nat_gateway_count

  domain = "vpc"
}

resource "aws_eip" "secondary" {
  count = local.nat_gateway_count * local.number_of_secondary_eips_per_gateway

  domain = "vpc"
}

resource "aws_nat_gateway" "this" {
  count = local.nat_gateway_count

  allocation_id = aws_eip.nat[count.index].id
  subnet_id     = local.public_subnets[count.index]

  secondary_allocation_ids = slice(aws_eip.secondary[*].id, count.index * local.number_of_secondary_eips_per_gateway, (count.index + 1) * local.number_of_secondary_eips_per_gateway)
}
```

I think this should be an issue on the aws provider repo if there is not one already.

velkovb: https://github.com/hashicorp/terraform-provider-aws/issues/33964

https://github.com/hashicorp/terraform-provider-aws/issues/33964

PR #1116, opened 2024-09-04
feat: Add outputs for the full list of subnets created and their attributes

certara-mchamberland:
## Description
This adds outputs mapping subnet IDs to AZs. This allows comparing AZs with a remote PrivateLink Endpoint Service to find out in what subnet we can create an endpoint locally, as PrivateLink requires both ends use the same AZs.

## Breaking Changes
No, these are only new outputs.

## How Has This Been Tested?
- [ ] I have updated at least one of the `examples/*` to demonstrate and validate my change(s) - Not sure its relevant for outputs
- [X] I have tested and validated these changes using one or more of the provided `examples/*` projects - I applied it directly to a module of mine that, among other things, locates matching AZs on both ends of a PrivateLink VPC Endpoint


certara-jgitlin:
LGTM! 👍

certara-mchamberland:
Could a maintainer pls take a look at this? I'm happy to keep working on it if its not currently adequate. Thanks!

outputs.tf
```diff
@@ -155,6 +155,11 @@ output "public_network_acl_arn" {
   value       = try(aws_network_acl.public[0].arn, null)
 }
 
+output "public_subnet_az_mapping" {
+  description = "A mapping of public subnet IDs to Availability Zones, useful when deploying PrivateLink endpoint services, which require matching AZs on both ends."
+  value       = { for subnet in aws_subnet.public[*] : (subnet["id"]) => subnet["availability_zone"] }
```
bryantbiggs, line 160: rather than this custom mapping, I think a better approach would be to just output all attributes of the subnets created and let users use those as they please (you can use this to create this mapping for example in your own implementation)

```hcl
output "public_subnets" {
  description = "List of the subnets created and their associate attributes"
  value       = aws_subnet.public
}
```

bryantbiggs (changes requested):

outputs.tf
certara-mchamberland: I'd built it this way to avoid a breaking change, hoping for a fast release, but I can arrange this too. Right now my org is working from my fork anyways. Alternatively, I could submit a second PR implementing the "breaking change" for merge in the next major version, your call!

outputs.tf
```diff
@@ -105,6 +105,11 @@ output "igw_arn" {
 # Publiс Subnets
 ################################################################################
 
+output "public_subnet_objects" {
+  description = "A list of all public subnets, containing the full objects."
+  value       = aws_subnet.public[*]
```
bryantbiggs, line 110: No need for the `[*]`, they can all be dropped if they are the last attribute selector
```suggestion
  value       = aws_subnet.public
```

Issue #978, opened 2023-08-28
This module, doesnt integrate with the Network Firewall Module


jseiser:
## Is your request related to a problem? Please describe.
Their is now a network firewall module, but it does not integrate with this module, so even when using them both, you are left in a situation where you can not actually route traffic through the firewall in its normal manner.

Ive also created this here: https://github.com/terraform-aws-modules/terraform-aws-network-firewall/issues/1

Since im not sure where it should live, since both modules are impacted

## Describe the solution you'd like.
1. Should be able to create `Firewall Subnets`
2. Should be able to control/override routes for the module so once the firewall is deployed, you can update the routing tables to point to it.

How we currently handle this

```

    Disable public subnets, and nat gateways
    Create Public Subnets and NAT Gateway, and Internet Gateway
    Create the route table for the NAT gateways
    Create 2 Firewall Subnets and their route tables
    Deploy this ( actually your old beta version) module passing in the firewall subnets in the subnet mapping
    Create the routes for the IGW
    Create the routes for the firewall subnets
    Create all the route associations
```

## Additional context
I think the main issue is not being flexible enough in subnet creation, and not being able to override routes from the Firewall Module, since anytime the VPC module is re-ran, they would be removed/correct I believe.

alekc:
If a customer doesn't want to use AWS nat gw (`enable_nat_gateway = false`) and use own solution, they cannot because the routing table is the same (or better, they cannot avoid cross AZ traffic and related cost). 

Is there advantage in keeping a single unified route table compared to one RT for every subnet?

dimakievukraine:
@antonbabenko hi. Is there any chance to look at this issue? I cannot use your Network Firewall module with this VPC module due to single and non flexible public subnet RT. 3 public subnet route tables + adding optional firewall VPC endpoint as route would help a lot.

bryantbiggs:
The network firewall module was designed [here](https://github.com/clowdhaus/terraform-aws-vpc-v5) which contains a different method for provisioning subnets. we are planning to continue to migrate that functionality over to the respective modules within this GitHub organization

kclinden:
Are there any branches that currently have this functionality?

blongman-snapdocs:
Starting on my journey to add Network Firewall to my own VPCs that use this module.  I'm hoping that whatever I do is easily convertible to use this module in the future.

euhn-gp:
[...]

PR #1127, opened 2024-11-03
feat: Add option to create/delete NAT Gateway route for private route tables

ixolt:
## Description
This update introduces a new variable, `create_private_nat_gateway_route`, to control the creation of NAT Gateway routes for private subnets. By making the creation of the `aws_route.private_nat_gateway` resource conditional, we provide users the flexibility to manage routes more precisely according to their specific requirements. 

**Key Changes:**
- **New Variable** - `create_private_nat_gateway_route` (Boolean)
  - When set to `true`(default): NAT Gateway routes for private subnets are created automatically.
  - When set to `false`: NAT Gateway routes are not created, allowing users to define routes manually.
- **Resource Modification** - The `aws_route.private_nat_gateway` resource now checks the value of `create_private_nat_gateway_route` before creating routes.
- **Use Case** 
  - Users can deploy NAT Gateways without setting automatic routes. This is particularly beneficial when a different routing strategy is in place, such as using a Transit Gateway for the default `0.0.0.0/0` route while assigning NAT Gateway routes only for specific destinations.

**Example Use Case:**
- If a Transit Gateway is used for the default `0.0.0.0/0` route and NAT Gateways are desired only for specific CIDR blocks, setting `create_private_nat_gateway_route` to `false` will prevent automatic route creation, giving full manual control over route configurations.

## Motivation and Context
The change is required to provide greater flexibility and control over the infrastructure setup. Previously, creating a NAT Gateway would always result in the automatic creation of routes, which limited configuration options for complex setups. By introducing this conditional variable, users can decide whether or not to create NAT Gateway routes, making the module more adaptable to different networking requirements.
This enhancement addresses scenarios where automatic route creation is not desirable or conflicts with specific routing strategies.
- https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/978

## Breaking Changes

No, this change does not break backward compatibility. The default behavior remains unchanged unless the `create_private_nat_gateway_route` variable is explicitly set to `false`.

## How Has This Been Tested?
- [ ] I have updated at least one of the `examples/*` to demonstrate and validate my change(s).
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects.
  - I deployed an example project with and without the `create_private_nat_gateway_route` variable set to ensure correct behavior.
  - Verified that the NAT Gateway routes are only deleted when the variable is disabled.
- [x] I have executed `pre-commit run -a` on my pull request to ensure code quality and compliance with standards.


antonbabenko (approved):

antonbabenko:
This PR is included in version 5.15.0 :tada:

PR #1138, opened 2024-11-18
feat: Added additional conditions into Flow Log IAM Role Assumption Policy

fschykowski:
## Description
Allows modifications to the assume role policy of the flow log IAM role. 

## Motivation and Context
The recommended action by AWS to prevent confused deputy attacks is to add sourceArn, sourceAccount or sourceOrg as conditions as condition to the role assumption. https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html

## Breaking Changes
Should not break anything, as the default variable is empty and will simply not render the dynamic block.

## How Has This Been Tested?
- [X] I have validated locally that the config is rendered correctly from terraform plan for both adding sourceAccount as condition and not adding conditions.


antonbabenko:
This PR is included in version 5.16.0 :tada:

PR #1140, opened 2024-11-19
chore(nat-gateway): Add nat_gateway_tags_per_az variable for NAT gateways

pavlospt:
## Description
Adding `nat_gateway_tags_per_az` variable for NAT gateways

## Motivation and Context
It is useful for cost related reports to be able to tag NAT gateways per AZ.

## Breaking Changes
No

## How Has This Been Tested?


pavlospt:
@bryantbiggs @antonbabenko Sorry for tagging you directly, but would it be possible to have a review here? 🙏🏻

bryantbiggs:
why is this a useful change, what motivates a specific tag per NAT?

pavlospt:
@bryantbiggs Extremely useful for cost observability reasons (at least in our case, we can know what volume of traffic is going through each of our NATs and correlated that with the workloads living there)!

bryantbiggs:
but you know the NAT gateway IDs and can use that to correlate, no? this sounds like a side effect of cross-AZ traffic which I don't see the benefit of taking on such complexity in this module for this peculiar case

pavlospt:
By using common tags across resources we can do what I describe above. I am not sure why or how this feature brings the complexity you mention, given that the info we need is already there and we just utilize it. 

Indeed it made things pretty clear for us regarding Cross AZ traffic costs. If you have better alternatives to suggest, we would love to investigate them!

Kretolus:
For us, this helps align the naming convention we're using for various AWS resources.

While for our use case all other resources we need can either have their name changed, or have relevant workarounds already in place in the module (subnet prefixes, tag overwrites, etc.), the NAT Gateways are the only one without an easy workaround.

It also helps with the billing mentioned above to make it more human readable than IDs.

pavlospt:
Not a stale PR

Issue #1150, opened 2024-12-17
VPC Endpoints should add Name tag by default


aldenquimby:
## Is your request related to a new offering from AWS?
No

## Is your request related to a problem? Please describe.
When adding vpc-endpoints, I currently add a tag for Name on each, like this:
```tf
    execute-api = {
      service = "execute-api"
      tags    = { Name = "execute-api-vpc-endpoint" }
    }
    textract-fips = {
      service = "textract-fips"
      tags    = { Name = "textract-fips-vpc-endpoint" }
    }
    transcribe = {
      service = "transcribe"
      tags    = { Name = "transcribe-vpc-endpoint" }
    }
```

## Describe the solution you'd like.
It would be nice if Name tags were auto-added if `service` is specified, so I could do this instead:
```tf
    execute-api = {
      service = "execute-api"
    }
    textract-fips = {
      service = "textract-fips"
    }
    transcribe = {
      service = "transcribe"
    }
```

## Describe alternatives you've considered.
Manually specifying Names works, it's just verbose

## Additional context
- The security group within the vpc-endpoints module already auto-adds a Name tag, which is nice:
https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/9ffd9c66f3d7eb4b5bc2d7bc7d049f794b127693/modules/vpc-endpoints/main.tf#L68-L72

sebastianczech:
@aldenquimby I prepared PR #1151 , which I hope resolves your problem

aldenquimby:
@sebastianczech excellent! Thank you. Exactly what I was hoping for

PR #1151, opened 2024-12-18
feat: Define default name for VPC endpoint

sebastianczech:
## Description

PR delivers feature for VPC endpoints - if name is not defined, then default name is constructed from service name.

## Motivation and Context

#1150 

## Breaking Changes

There are no breaking changes.

## How Has This Been Tested?

### DIFF
--- a/modules/vpc-endpoints/main.tf
+++ b/modules/vpc-endpoints/main.tf
@@ -46,7 +46,7 @@ resource "aws_vpc_endpoint" "this" {
 
   tags = merge(
     var.tags,
-    { "Name" = "${each.key}-vpc-endpoint" },
+    { "Name" = replace(try(each.value.service_endpoint, data.aws_vpc_endpoint_service.this[each.key].service_name, ""), ".", "-"") },
     try(each.value.tags, {}),
   )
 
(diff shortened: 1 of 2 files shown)


modules/vpc-endpoints/main.tf
```diff
@@ -44,7 +44,11 @@ resource "aws_vpc_endpoint" "this" {
     }
   }
 
-  tags = merge(var.tags, try(each.value.tags, {}))
+  tags = merge(
+    var.tags,
+    { "Name" = "${each.key}-vpc-endpoint" },
```
bryantbiggs, line 49:
```suggestion
    { "Name" = replace(try(each.value.service_endpoint, data.aws_vpc_endpoint_service.this[each.key].service_name, ""), ".", "-"") },
```

sebastianczech: Thanks @bryantbiggs for quick suggestions, I agree that it's better approach to use service name.

bryantbiggs: meh, I deployed that and forgot the `com.amazonaws.<region>` is in there so reverted back to your original suggestion

bryantbiggs (approved):

modules/vpc-endpoints/main.tf
sebastianczech: At the beginning I was also adding suffix `-vpc-endpoint` . Can I restore it too ?

antonbabenko:
This PR is included in version 5.17.0 :tada:

Issue #1153, opened 2024-12-30
Ability to configure IPAM exclusively for IPv4 or IPv6


Vinoth-S15:
## Is your request related to a problem? Please describe.
The current implementation of the aws_vpc resource does not provide the flexibility to independently use an IPAM pool for IPv4 and IPv6. In cases where IPAM is needed exclusively for IPv6 CIDR allocation, the configuration becomes inefficient and restrictive

## Describe the solution you'd like.
To enhance flexibility, the assign_generated_ipv6_cidr_block argument should be updated to use the condition var.ipv6_ipam_pool_id == null instead of !var.use_ipam_pool. This change would allow IPAM to be configured specifically for IPv6 without impacting IPv4.

## Additional context
<img width="1549" alt="image" src="https://github.com/user-attachments/assets/131b540c-aff0-4e08-bb50-2a051f12e8b2" />

sebastianczech:
@Vinoth-S15 I prepare #1155 with changes in module and example. Are you able to take a quick look , if it's ok for you , please ?

Vinoth-S15:
@sebastianczech The changes are satisfactory to me.

sebastianczech:
@bryantbiggs , are you able to take a look on this PR , please ?

Vinoth-S15:
@bryantbiggs are you able to have a look on this PR #1155, please ?

Vinoth-S15:
@bryantbiggs are you able to have a look on this PR https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/1155, please ?

Vinoth-S15:
.

iaacautomation:
AWS does not allow ipv6 without ipv4 so you can't have an exclusive ipv6 network, if you want ipv4 only, set enable_ipv6 to false

sebastianczech:
@iaacautomation as far as I understand this is not about using only IPv6 in AWS, rather about using IPAM exclusively for IPv6 and still using both IPv4 and IPv6 in AWS.

PR #1155, opened 2025-01-04
feat!: Add ability to configure IPAM exclusively for IPv4 or IPv6

sebastianczech:
## Description

PR delivers features to configure IPAM exclusively for IPv4 or IPv6 - previously it was 1 variable used for both IP protocols, so IPAM could be used for both or neither. 

## Motivation and Context

#1153 

## Breaking Changes

Previously it was defined 1 variable `use_ipam_pool` used for both IP protocol. In order to distinguish to which protocol IPAM should be used, there were introduced 2 new variables `use_ipv4_ipam_pool` and `use_ipv6_ipam_pool` (old variable   `use_ipam_pool` was removed).

## How Has This Been Tested?


Vinoth-S15:
The changes are satisfactory to me

bryantbiggs:
you need to use `enable_ipv6` to toggle on IPv6 - I don't know that these breaking changes are warranted at this time

Vinoth-S15:
We might need IPAM only for IPv6 and not for IPv4. Enabling IPAM for IPv4 will monitor all the private IPs it manages, which adds more cost.

Ex:
For 5,000 private IPs managed by AWS IPAM, assuming all are active, the monthly cost can be estimated as follows.

    Hourly Cost per IP: $0.00027 (this rate may vary by Region)
    Total Hourly Cost for 5,000 IPs: 5,000 IPs × $0.00027/IP = $1.35
    Monthly Cost: $1.35/hour × 24 hours/day × 30 days/month = $972

PR #1161, opened 2025-01-27
feat: Support for cross region VPC endpoints

mmeidlinger:
## Description
Adds cross-region capabilities for VPC endpoints.

## Motivation and Context
AWS added cross-region support for VPC Private link in November 2024. Support for that feature was added in the AWS [Terraform provider release 5.83](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.83.0). 

## Breaking Changes
No breaking changes, backwards compatible.

## How Has This Been Tested?
Tested and validated both the old syntax (of not providing the `service_region` as part of `var.endpoints` as well as with adding it.

### DIFF
--- a/modules/vpc-endpoints/README.md
+++ b/modules/vpc-endpoints/README.md
@@ -56,13 +56,13 @@ module "endpoints" {
 | Name | Version |
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.0 |
-| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 5.46 |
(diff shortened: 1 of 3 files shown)


antonbabenko (approved):

antonbabenko:
This PR is included in version 5.18.0 :tada:

mmeidlinger:
Thanks, awesome turnaround time @antonbabenko
I think though I missed one case though to make it more robust. When we explicitly pass `service_region = null`, then we get an error because I pass on `[null]` rather than `null`. Changing this line to
```hlc
service_regions = each.value.service_region != null ? [each.value.service_region] : null
```
should fix that.

PR #1162, opened 2025-01-27
fix: Fixed service_region argument in the VPC endpoint module

mmeidlinger:
## Description
Currently, if we pass ` service_region=null` (rather than not passing it at all), the module fails as it passes on `[null]` rather than `null` to the underlying data source.

## Motivation and Context
See description and [the comment in this PR that added cross-region support](https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/1161#issuecomment-2616115424) in the first place.

## Breaking Changes
Non breaking, makes things more robust and covers all all cases (no `service_region` passed, `service_region=null` passed, valid service region passed)

## How Has This Been Tested?
Tested passing no `service_region`, `service_region=null`, or a valid service region passed

### DIFF
--- a/modules/vpc-endpoints/main.tf
+++ b/modules/vpc-endpoints/main.tf
@@ -13,7 +13,7 @@ data "aws_vpc_endpoint_service" "this" {
 
   service         = try(each.value.service, null)
   service_name    = try(each.value.service_name, null)
-  service_regions = try([each.value.service_region], null)
+  service_regions = try(each.value.service_region != null ? [each.value.service_region] : null, null)
 
   filter {
     name   = "service-type"
--- a/modules/vpc-endpoints/main.tf
+++ b/modules/vpc-endpoints/main.tf
@@ -13,7 +13,7 @@ data "aws_vpc_endpoint_service" "this" {
 
   service         = try(each.value.service, null)
   service_name    = try(each.value.service_name, null)
-  service_regions = try(each.value.service_region != null ? [each.value.service_region] : null, null)
+  service_regions = try(each.value.service_region, null) != null ? [each.value.service_region] : null
 
   filter {
     name   = "service-type"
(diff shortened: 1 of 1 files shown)


modules/vpc-endpoints/main.tf
```diff
@@ -13,7 +13,7 @@ data "aws_vpc_endpoint_service" "this" {
 
   service         = try(each.value.service, null)
   service_name    = try(each.value.service_name, null)
-  service_regions = try([each.value.service_region], null)
+  service_regions = try(each.value.service_region != null ? [each.value.service_region] : null, null)
```
antonbabenko, line 16:
```suggestion
  service_regions = one([each.value.service_region])
```

Much simpler, but please test it. :)

antonbabenko (changes requested):

modules/vpc-endpoints/main.tf
mmeidlinger: Thanks for the feedback. So I tested 
```hcl
service_regions = one([each.value.service_region])
```
this does not quite work. If  the `endpoints` maps' `service_region` value is a valid string, it evaluates just to the string, but the data sources' `service_regions` expects a list of strings. Further, if `service _region` is not included in the `endpoints` map, it also fails.

mmeidlinger: I think we could change it to this which is a bit easier to understand and read and I successfully tested for all 3 cases (`service_region` not included in the `endpoints` map,  `service_region=null` or  `service_region="valid-region-string"`:
```hcl
service_regions = try(each.value.service_region, null) != null ? [each.value.service_region] : null
```

antonbabenko: I think this is going to work as expected:

```hcl
service_regions = try(coalescelist(compact([each.value.service_region])), null)
```

Let's use it if it does.

mmeidlinger: It does indeed, thank you! Updated the PR accordingly.

antonbabenko (approved):

modules/vpc-endpoints/main.tf
antonbabenko: Thank you for fixing an issue!

PR #1186, opened 2025-05-02
fix: Update aws_vpn_gateway_route_propagation private/intra count to …

GrahamOHagan:
## Description
Update the `"aws_vpn_gateway_route_propagation" "private"/"intra"` resources to count the route tables instead of the subnet.

## Motivation and Context
Where there is a single route table and three subnets, e.g. when `"single_nat_gateway" = true`, the same propagation resource becomes managed multiple times under this module.

When switching an existing VPC from a single NAT gateway to three NAT gateways, terraform recreates the [1] and [2] under the new route tables, this removes the route from the original route table.

## Breaking Changes
If already using `"single_nat_gateway" = true`, two terraform applies are required. One to delete the additional propagations, and another to recreate the [0] propagation. Alternatively run terraform state rm on the extra propagations.

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1219,8 +1219,9 @@ resource "aws_vpn_gateway_route_propagation" "public" {
   )
 }
 
+# There are as many routing tables as the number of NAT gateways
 resource "aws_vpn_gateway_route_propagation" "private" {
-  count = local.create_vpc && var.propagate_private_route_tables_vgw && (var.enable_vpn_gateway || var.vpn_gateway_id != "") ? local.len_private_subnets : 0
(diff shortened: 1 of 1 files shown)


bryantbiggs:
this is not a valid change from what I can understand. it appears to be an issue with adopting and existing VPC created outside this module, which is not something we design for

GrahamOHagan:
> this is not a valid change from what I can understand. it appears to be an issue with adopting and existing VPC created outside this module, which is not something we design for

Hi, this is related to VPC resources created & managed by this module. The issue is one resource is managed multiple times by Terraform.

PR #1188, opened 2025-05-12
feat: Add toggle to disable default route creation for public route tables

### LINKED ISSUE #1187: Feature to disable default 0.0.0.0/0 route for public subnets

This feature is not bound to a specific AWS feature(other than used by the VPC module already)

When using network firewall we need to be able to have a default route through the network firewall for our public subnets

As the VPC today always creates the 0.0.0.0/0 route (and also creates the route table for the public subnets). We have no way to use the this VPC module with a network firewall

I think the simplest solution with least change is to be able to disable the creation of the default route for public subnets

The alternative I considered first was having a toggle to disable creating the route table(s) for public subnets and then provide "my own" route table(s)

## Additional context

Would this be an ok approach?

HasseJohansen:
## Description

This changes introduces a variable public_enable_default_route so a default route for public subnets is not created when set to false . This enables the user to create its own default route to other gateways than Internet Gateway (IGW)

## Motivation and Context

It fixes the problem of not being able to use ex. Network Firewall and enables the use case where users will need to change default route to something else than the IGW

## How Has This Been Tested?

I have tested my changes from a module using the VPC. Setting the public_enable_default_route  flag to false -> default route is not created for public subnet route tables

removing the public_enable_default_route flag from the settings passed to the module -> The default route to IGW is created for each public subnet route table

closes: #1187


HasseJohansen:
Please do not close. Can we have some action on this? or a decission if this is going to be merged or  you would merge https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/1190 ?

HasseJohansen:
Please do not close

bryantbiggs:
I'll take a look at the Network Firewall integration options today

bengaywins:
@bryantbiggs Is there any idea of if/when this would get merged? We are deploying an NGFW and without this, we're basically forcing the allowance of drift. Which isn't ideal for obvious reasons, but this is the big thing we need to allow for changing the 0/0 route.

(yes I know I posted this same thing in both but it's not clear which may or may not be merged)

HasseJohansen:
I would also like that this gets some traction. I must say I am surprised that there is not more activity in what I believe is the goto module for vpc creation in the terraform community

I have been using my fork since I opened this pull request - also in production

and anyway it is a _very_ small change which defaults to the “old behaviour” so I don’t see much risk in merging it (but maybe maintainers see something I haven’t though about)

bryantbiggs:
> I would also like that this gets some traction. I must say I am surprised that there is not more activity in what I believe is the goto module for vpc creation in the terraform community
> 
> I have been using my fork since I opened this pull request - also in production
> 
> and anyway it is a _very_ small change which defaults to the “old behaviour” so I don’t see much risk in merging it (but maybe maintainers see something I haven’t though about)

folks tend to look at changes in isolation, which we do not (nor could we, otherwise the modules wouldn't be as widely used)

we also have to work at jobs that generate income since these modules provide zero income. so if you wish to gain more traction on changes, consider sponsoring us. I think people would s*&t their pants if they knew how much revenue is attributed to these modules, but we don't see a dime of it (revenue on the AWS side as well as the consumer/user side)

Chili-Man:
> folks tend to look at changes in isolation, which we do not (nor could we, otherwise the modules wouldn't be as widely used)
> 
> we also have to work at jobs that generate income since these modules provide zero income. so if you wish to gain more traction on changes, consider sponsoring us. I think people would s*&t their pants if they knew how much revenue is attributed to these modules, but we don't see a dime of it (revenue on the AWS side as well as the consumer/user side)

@bryantbiggs You have the community here at your disposal willing to help - share that burden with us.  I've been trying to get the companies I work at to sponsor this project, but the reality is its difficult to convince leadership especially when we're startups and they respond by saying to just fork the project, which unfortunately is what we end up doing.  I can personally donate my time to help out, so what can I do specifically to help move this feature forward?  If there's a bigger picture going on here, then how can I help move that forward?  What do we need to do?

bryantbiggs:
thank you but it usually just ends up coming down to one of the maintainers doing the work when things get tricky/difficult - case in point
- https://github.com/terraform-aws-modules/terraform-aws-transit-gateway/pull/113#issuecomment-3775166144
- https://github.com/terraform-aws-modules/terraform-aws-transit-gateway/pull/153

so while volunteering sounds great, we haven't seen much success with that. did you know there 16 folks in the org? more than just the few you see who respond and work on issues/PRs. same thing - folks who said they would do something and they never do. so while I appreciate the offer, the truth is that if anything is going to get done, its most likely going to come from myself or maybe 1 other person 
image

spencercsheehan1:
Couple of small naming nits, in case it helps move this along ever in the future (totally understand there are limited maintainers - in case they decide to look at this bug in the future, otherwise feel free to ignore)!

**1.** The variable is an enable flag (`default = true`), but the description reads like a disable:

```hcl
description = "Disable default route to internet gateway for public subnets"
```

In the generated README that renders as "Disable default route..." next to a default of `true`, which reads backwards. Something like `"Controls if the default route to the internet gateway should be created for the public subnets"` would match the polarity, and matches how `create_private_nat_gateway_route` is worded.

**2.** The README snippet uses a name that isn't in the PR:

```hcl
public_disable_default_route = false
```

Should be `public_enable_default_route` - the prose just above it gets it right.

HasseJohansen:
@spencercsheehan1 Thank you for the comments. The bool was actually switched when I started, but I quickly realised that I could make the terraform code a little simpler flipping the switch. Somewhat stupid that I did not got that reflected in all places in the README

It should be correct now

Issue #1193, opened 2025-05-27
Can I create extra private subnets without tagging them?


Israphel:
## Is your request related to a new offering from AWS?
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - Yes ✅: please list the AWS provider version which introduced this functionality

## Is your request related to a problem? Please describe.
We use EKS, and there's a feature to use a secondary CIDR block only for the pods. In order to do that, you create a secondary CIDR block and extra private subnets associated to that. All of that is supported by this module.
However, for EKS users it's typical to have private subnet tags related to AWS load balancer and/or Karpenter. But, the subnets which are specific for the pods can't be tagged with those values, or they will be autodiscovered and used for something they shouldn't.

## Describe the solution you'd like.
Someway to create "extra" private subnets but without being affected by the private subnet tags.

## Describe alternatives you've considered.
I made a small module that creates subnets and route table associations, so I'm managing my vpc with this module plus that one.

PR #1139, opened 2024-11-19
feat: Support defining tags per subnet name

fatmcgav:
## Description

This commit updates the module to support defining subnet tags per subnet name, in addition to per AZ.

## Motivation and context

Reasoning behind this is to allow for specific tags to be applied to different public/private subnets.
E.g.
```terraform
  private_subnet_names = ["Private Subnet One", "Private Subnet Two"]
  private_subnet_tags_per_name = {
    "Private Subnet One" = {
      "subnet one" = "true"
    },
    "Private Subnet Two" = {
      "subnet two" = "true"
    }
  }
```

Fixes #1129

PR #1217, opened 2025-06-30
feat: Issue 1193 support for tagging subnets by name

voidshard:
## Description

Allows one to add per-subnet tags, settable by subnet name.

## Motivation and Context
Currently it is possible to set tags on all subnets, or all subnets of a given distinction (eg. public, private, ..) but it is not possible to set tags on a given subnet. This is the motivation behind [this issue](https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/1193) and the related [MR](https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/1139/files). This MR adds the same support for each subnet type rather than just public & private. 

## Breaking Changes
None

## How Has This Been Tested?

I added my own simple example. It includes cases where all, none and a subset of subnets have custom tags added by name.


voidshard:
@antonbabenko @bryantbiggs please let me know your thoughts - we switched to using this module at work to define our VPCs but need a bit more flexibility to tag subnets appropriately for autodiscovery & configuration purposes.

jpriebe:
I can't believe how resistant the maintainers have been to this concept.  Not even just the implementations -- it seems like the very concept is anathema to them.

We have multiple sets of private subnets.  Only one set should be tagged with EKS tags.  But if we tag via terraform outsidde of this module, the module wants to remove them on the next apply, and then on the next apply after that, our terraform wants to restore them.  I'm stuck with infinite noise and potential breakage of EKS.

Please provide a viable workaround, or at least tell us this feature is on your short-term roadmap.

bryantbiggs:
> I can't believe how resistant the maintainers have been to this concept.  Not even just the implementations -- it seems like the very concept is anathema to them.
> 
> 
> 
> We have multiple sets of private subnets.  Only one set should be tagged with EKS tags.  But if we tag via terraform outsidde of this module, the module wants to remove them on the next apply, and then on the next apply after that, our terraform wants to restore them.  I'm stuck with infinite noise and potential breakage of EKS.
> 
> 
> 
> Please provide a viable workaround, or at least tell us this feature is on your short-term roadmap.

Please provide funding, or at least have patience as we work through a number of significant changes

voidshard:
If you want to ask for funding you first have to clearly communicate your plans and intentions no one provides funding into an unknown black box. Well, that isn't a government.

PR #1231, opened 2025-08-06
fix: security_group_tags for endpoints

viktor-kanavchenko:
## Description
Rearrange Security Group tags for Endpoints. Position generated Name tag before user input.

## Motivation and Context
This change fix a bug that makes impossible to define custom Name for the SG.

## Breaking Changes
none

## How Has This Been Tested?

### DIFF
--- a/modules/vpc-endpoints/main.tf
+++ b/modules/vpc-endpoints/main.tf
@@ -82,9 +82,9 @@ resource "aws_security_group" "this" {
   vpc_id      = var.vpc_id
 
   tags = merge(
+    { "Name" = try(coalesce(var.security_group_name, var.security_group_name_prefix), "") },
     var.tags,
     var.security_group_tags,
-    { "Name" = try(coalesce(var.security_group_name, var.security_group_name_prefix), "") },
   )
 
   lifecycle {


viktor-kanavchenko:
@bryantbiggs since you close the PR, can you please help me find a way how to set a custom Name tag for Endpoint SG?

The root module from this repo has the same pattern of forming resources tags: [example1](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/main.tf#L46-L50), [example2](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/main.tf#L46-L50), [example3](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/main.tf#L147-L157), [example4](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/main.tf#L147-L157), [example5](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/main.tf#L218-L222), [example6](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/main.tf#L301-L310). The user provided tags have precedence over generated ones.

bryantbiggs:
Why do you need to set a custom name tag on and endpoint?

viktor-kanavchenko:
I'm sorry if I miss led you. The Name tag is needed for endpoints' Security Group. Please do not mix up with _Security Group Name_.
My use case - I do migration of CloudFormation template to Terraform configuration. The Name tag is already set, so I want to reflect it in my configuration. Because of the argument order in `merge()` function, an autogenerated value always overrides user input, and as a result terraform plan shows changes.
Another case might be do not set Name tag(to pass empty string or `null`).

Issue #1234, opened 2025-08-12
Unable to customise vpc endpoint security group Name tag


chris3ware:
## Description

When supplying `security_group_name_prefix` as an input to the vpc endpoint module, with a trailing hyphen to separate the user defined prefix from the random suffix, the `Name` tag ends with the trailing hyphen.

To attempt to overcome this, I have tried to pass in a `Name` tag via `security_group_tags`. This does not work due to the precedence of the `merge` function on the [`aws_security_group`](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/0eefe64aab5bf762b06de83a6219d417a56a489c/modules/vpc-endpoints/main.tf#L87) resource preferring the hardcoded `Name` tag as it is the later argument in the sequence.

If this is compared to the [ec2-instance module](https://github.com/terraform-aws-modules/terraform-aws-ec2-instance/blob/da57ed454d7ef513b0a5dd4f95421c593bb4d4b0/main.tf#L754) - the `merge` function for security group tags allows a user to pass in a custom name because `var.security_group_tags` is last in the sequence.

I raised https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/1233 to provide a fix for this but it has been closed.https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/1231 has also been closed for the same issue.

For clarity: the requirement is to be able to specify a custom `Name` tag for the VPC endpoint security group, not the endpoint itself.

## Versions

- Module version [Required]: v6.0.1

- Terraform version: v1.5.7

- Provider version(s):  provider registry.terraform.io/hashicorp/aws v6.8.0

## Reproduction Code [Required]

```hcl
provider "aws" {
  region = local.region
}

data "aws_availability_zones" "available" {}

locals {
  name   = "ex-${basename(path.cwd)}"
  region = "eu-west-1"

  vpc_cidr = "10.0.0.0/16"
  azs      = slice(data.aws_availability_zones.available.names, 0, 3)

  tags = {
    Example    = local.name
    GithubRepo = "terraform-aws-vpc"
    GithubOrg  = "terraform-aws-modules"
  }
}

module "vpc" {
  source = "../../"

  name = local.name
  cidr = local.vpc_cidr

  azs                 = local.azs
  private_subnets     = [for k, v in local.azs : cidrsubnet(local.vpc_cidr, 8, k)]
}

module "vpc_endpoints" {
  source = "../../modules/vpc-endpoints"

[...]

PR #1235, opened 2025-08-12
fix: Prefer user defined `security_group_tags` in the vpc endpoint module

chris3ware:
## Description
User defined tags, which could include a `Name`, for the VPC endpoint security group should be preferred over hardcoded tags. This aligns the functionality with other resources in this and other terraform-aws-modules; several examples are provided in https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/1231

## Motivation and Context
Resolves: https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/1234

## How Has This Been Tested?

### DIFF
--- a/examples/complete/main.tf
+++ b/examples/complete/main.tf
@@ -102,6 +102,10 @@ module "vpc_endpoints" {
     }
   }
 
+  security_group_tags = {
+    Name = "example-name-override"
+  }
+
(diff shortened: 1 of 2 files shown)


bryantbiggs:
it is correct as it is currently written so this change is not warranted

PR #1237, opened 2025-08-19
feat: Add per AZ tags support for NAT Gateway, NAT EIP and private route table

aayushsrivastava:
## Description

Adds the ability to customize the tags for an individual NAT Gateway, NAT EIP, or private route table.

## Motivation and Context
The components can have multiple instances in different AZs.

We are trying to import some existing VPC resources into Terraform. These resources have some tags that differ in value between AZs, and would like the Terraform module to be able to accomodate them.

## Breaking Changes
No breaking changes

## How Has This Been Tested?
- [x] I have tested and validated these changes on an internal Terraform configuration at our company.


bryantbiggs:
thank you but this is not something we are going to accept at this time - tags should be standardized across resources and not so many unique/snowflake tags

Issue #1239, opened 2025-09-16
Custom names for part of the subnets


RafaelWO:
## Is your request related to a new offering from AWS?
No

## Is your request related to a problem? Please describe.
I want to give a name to some of my private subnets (via `private_subnet_names`) but not for all of them. It would be helpful if I could set a value, e.g. `null`, for this module to do the naming and pass my name otherwise.

For example:
```tf
private_subnets      = ["10.0.32.0/22",  "10.0.36.0/22"]
private_subnet_names = [null, "my-net"]
```

The variables above should result in the first subnet name being automatically assigned, while the second one is `my-net`.

## Describe the solution you'd like.
Currently, the `Name` tag is either picked from `var.private_subnet_names[count.index]` or named automatically, if the indexing fails (via `try()`).

We could change this logic to:
```diff
-      Name = try(
-        var.private_subnet_names[count.index],
-        format("${var.name}-${var.private_subnet_suffix}-%s", element(var.azs, count.index))
+      Name = (
+        try(var.private_subnet_names[count.index], null) != null
+        ? var.private_subnet_names[count.index]
+        : format("${var.name}-${var.private_subnet_suffix}-%s", element(var.azs, count.index))
      )
```

## Describe alternatives you've considered.
One could also do the logic described above outside of this module. But one could not rely on the autogeneration of names staying in sync.

## Additional context
I am happy to create a PR with the changes described - or the adjusted logic after discussions with the maintainer 🙂

shivamchaubey027:
Hi @RafaelWO, I'd like to implement the solution you've outlined here. Would you prefer to work on this yourself, or would you be okay with me taking it on?

bryantbiggs:
thank you but we are not going to make any further changes to the various custom naming schemes of subnets on the current module structure

PR #1240, opened 2025-09-16
feat: Allow partial subnet naming

shivamchaubey027:
Implements the solution proposed by @RafaelWO in issue #1239.

## Description

This change updates the naming logic for private subnets to allow for partial custom naming. It enables the use of `null` values in the `private_subnet_names` list to fall back to the module's default auto-naming convention for specific subnets.

## How Has This Been Tested?

- Tested locally using the example test case with `private_subnet_names = [null, "my-custom-net-2", null]`.
- Confirmed that `terraform plan` produces the expected names for all three subnets (one custom, two auto-generated) and resolves the original issue.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -282,9 +282,9 @@ resource "aws_subnet" "private" {
   tags = merge(
     {
       Name = (
-       try(var.private_subnet_names[count.index], null) != null
-        ? var.private_subnet_names[count.index]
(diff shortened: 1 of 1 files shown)


bryantbiggs:
thank you but we are not going to make any further changes to the various custom naming schemes of subnets on the current module structure

PR #1241, opened 2025-09-25
feat: Add new sub-module for AWS flow log

bryantbiggs:
## Description
- Add new sub-module for AWS flow log

## Motivation and Context
- To support more scenarios
	- Multiple flow logs per VPC
	- Flow logs outside of this module (https://github.com/terraform-aws-modules/terraform-aws-transit-gateway/pull/145)
- This a precursory change that is valid in this modules current form. At the next breaking change the embedded flow log implementation will be removed and users migrated to the sub-module instead. Adding this now reduces the size/scope of that later PR

## Breaking Changes
- No, all changes are self-contained to this new sub-module only. The example as well as example outputs have been removed to start dissuading users from using the current implementation and look to use this sub-module instead. However, current functionality remains intact and untouched 

## How Has This Been Tested?


modules/flow-log/versions.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = ">= 1.5.7"
```
bryantbiggs, line 2: these versions are different from the rest of the repository but that is intention. It does not affect anything else in the module, but we want to start off on these versions

PR #1274, opened 2026-01-07
fix: Fix availability zone regex conditions to work for European sovereign cloud

bpauwels:
Global Partition:
```
~ $ aws ec2 describe-availability-zones                                                                                                                                                                                                                                                                                                                
{
    "AvailabilityZones": [
        {
            "OptInStatus": "opt-in-not-required",
            "Messages": [],
            "RegionName": "us-west-2",
            "ZoneName": "us-west-2a",
            "ZoneId": "usw2-az1",
            "GroupName": "us-west-2-zg-1",
            "NetworkBorderGroup": "us-west-2",
            "ZoneType": "availability-zone",
            "GroupLongName": "US West (Oregon) 1",
            "State": "available"
        },
...
```

European Sovereign Cloud:
```
~ $ aws ec2 describe-availability-zones                                                                                                                                                                                                                                                                                                                
{
    "AvailabilityZones": [
        {
            "OptInStatus": "opt-in-not-required",
            "Messages": [],
            "RegionName": "eusc-de-east-1",
            "ZoneName": "eusc-de-east-1a",
            "ZoneId": "euscdee1-az1",
            "GroupName": "eusc-de-east-1-zg-1",
            "NetworkBorderGroup": "eusc-de-east-1",
            "ZoneType": "availability-zone",
            "GroupLongName": "EU (Germany) 1",
            "State": "available"
        },
...
```

Updated the regex to work for both formats of ZoneName / ZoneId

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -148,8 +148,8 @@ resource "aws_subnet" "public" {
   region = var.region
 
   assign_ipv6_address_on_creation                = var.enable_ipv6 && var.public_subnet_ipv6_native ? true : var.public_subnet_assign_ipv6_address_on_creation
-  availability_zone                              = length(regexall("^[a-z]{2}-", element(var.azs, count.index))) > 0 ? element(var.azs, count.index) : null
-  availability_zone_id                           = length(regexall("^[a-z]{2}-", element(var.azs, count.index))) == 0 ? element(var.azs, count.index) : null
+  availability_zone                              = length(regexall("-az[0-9]+$", element(var.azs, count.index))) == 0 ? element(var.azs, count.index) : null
+  availability_zone_id                           = length(regexall("-az[0-9]+$", element(var.azs, count.index))) > 0 ? element(var.azs, count.index) : null
   cidr_block                                     = var.public_subnet_ipv6_native ? null : element(concat(var.public_subnets, [""]), count.index)
   enable_dns64                                   = var.enable_ipv6 && var.public_subnet_enable_dns64
   enable_resource_name_dns_aaaa_record_on_launch = var.enable_ipv6 && var.public_subnet_enable_resource_name_dns_aaaa_record_on_launch
--- a/main.tf
+++ b/main.tf
@@ -296,8 +296,8 @@ resource "aws_subnet" "private" {
   region = var.region
 
   assign_ipv6_address_on_creation                = var.enable_ipv6 && var.private_subnet_ipv6_native ? true : var.private_subnet_assign_ipv6_address_on_creation
-  availability_zone                              = length(regexall("^[a-z]{2}-", element(var.azs, count.index))) > 0 ? element(var.azs, count.index) : null
-  availability_zone_id                           = length(regexall("^[a-z]{2}-", element(var.azs, count.index))) == 0 ? element(var.azs, count.index) : null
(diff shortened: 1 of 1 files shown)


antonbabenko (changes requested):
Hi Benedikt! 👋 

It looks valid to me, but could you please update it in all other places in this module as well?

bpauwels:
Hi Anton,

sure, I was missing the subnet groups!

antonbabenko (approved):
LGTM

@bryantbiggs WDYT?

main.tf
```diff
@@ -148,8 +148,8 @@ resource "aws_subnet" "public" {
   region = var.region
 
   assign_ipv6_address_on_creation                = var.enable_ipv6 && var.public_subnet_ipv6_native ? true : var.public_subnet_assign_ipv6_address_on_creation
-  availability_zone                              = length(regexall("^[a-z]{2}-", element(var.azs, count.index))) > 0 ? element(var.azs, count.index) : null
-  availability_zone_id                           = length(regexall("^[a-z]{2}-", element(var.azs, count.index))) == 0 ? element(var.azs, count.index) : null
+  availability_zone                              = length(regexall("-az[0-9]+$", element(var.azs, count.index))) == 0 ? element(var.azs, count.index) : null
```
bryantbiggs, line 151: I don't think this regex is correct - #370 introduced this logic to determine if zone ID or name should be used which I think we can improve without this logic. so for now lets
1. Add a `TODO` to remove this in the next breaking change and instead be more explicit (users either pass `availability_zone_name` or `availability_zone_id`, we probably don't want to use `availability_zone` since its intent is now ambiguous and will probably lead to confusing errors due to the change)
2. Lets change this logic to simplify it. A zone ID will only have two parts separated by a hyphen, zone names will have more parts. So lets do:
```suggestion
  availability_zone                              = length(split("-", element(var.azs, count.index))) > 1 ? element(var.azs, count.index) : null
```

main.tf
```diff
@@ -148,8 +148,8 @@ resource "aws_subnet" "public" {
...
+  availability_zone_id                           = length(regexall("-az[0-9]+$", element(var.azs, count.index))) > 0 ? element(var.azs, count.index) : null
```
bryantbiggs, line 152:
```suggestion
  availability_zone_id                           = length(split("-", element(var.azs, count.index))) == 1 ? element(var.azs, count.index) : null
```

main.tf
antonbabenko: I think the point of this PR is mostly to allow this module to work with the upcoming EU partition, where AZ properties are named differently, and the module is not properly recognizing them. (I didn't run the code for real, but only in my head)

antonbabenko:
@bpauwels Congrats on the launch! Will you be able to address the comments in this PR?

bryantbiggs:
if it works as its written, I am fine with it. I just think trying to match on structure will be more stable than trying to match on content (since content will change, highly unlikely structure will though)

kgoutham93:
@antonbabenko shall I take forward changes proposed by @bryantbiggs? This is biting us, as EUSC az names are passed as az IDs due to the regex check failure.

antonbabenko:
@kgoutham93 yes, please fix it.

kgoutham93:
Dropped a PR #1279 addressing @bryantbiggs review comments

antonbabenko:
This PR is included in version 6.7.1 :tada:

PR #1287, opened 2026-03-10
feat: Allow Amazon-provided IPv6 CIDR even if use_ipam_pool=true

alisson276:
## Description

If we have use_ipv6 enabled, the main logic of the use_ipam_pool is all-or-nothing, or we use IPAM for all, or the VPC fail to allocate IPv6 if no pool is provided.

Before:
- Enabling `use_ipam_pool=true` (for IPv4 IPAM) implicitly disabled Amazon-provided IPv6 CIDR assignment.

After:
- `assign_generated_ipv6_cidr_block` is enabled when:
  - `enable_ipv6 = true`
  - `ipv6_ipam_pool_id` is `null`
  - `ipv6_cidr` is `null`

This allows using IPv4 IPAM and AWS-provided IPv6 CIDR independently.

Also did a small change in the README to use `resource` instead of data block, because data block doesn't persist across multiple terraform plans, and try to recreate the subnets with another CIDR range everytime.

## Motivation and Context

This change fixes an issue where setting `use_ipam_pool=true` forced IPv6 IPAM behavior as well, even when the intent was to use the Amazon-provided IPv6 pool.

Problem solved:
- Decouples IPv4 IPAM selection from IPv6 CIDR source selection.
- Supports the valid combination: IPv4 from IPAM + IPv6 from Amazon pool.

## Breaking Changes

No.

## How Has This Been Tested?

I have tested the changes locally with my current IPAM installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ data "aws_vpc_ipam_pool" "ipv4_example" {
 }
 
 # Preview next CIDR from pool
-data "aws_vpc_ipam_preview_next_cidr" "previewed_cidr" {
+resource "aws_vpc_ipam_preview_next_cidr" "previewed_cidr" {
   ipam_pool_id   = data.aws_vpc_ipam_pool.ipv4_example.id
   netmask_length = 24
 }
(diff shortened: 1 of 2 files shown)


alisson276:
Is this doable or not?

alisson276:
@antonbabenko , could you please at least review the PR before it closes?

PR #1288, opened 2026-03-11
feat: GatewayLoadBalancer supported with subnet_ids

codyrat:
## Description
feat: Allow `subnet_ids` to be passed for `GatewayLoadBalancer` endpoint type

The `subnet_ids` argument on `aws_vpc_endpoint` is valid for both `Interface` and `GatewayLoadBalancer` endpoint types, but the module currently only passes `subnet_ids` when `service_type` is `Interface`. 

This PR updates the `modules/vpc-endpoints/main.tf` to include `GatewayLoadBalancer` in the list of types that receive `subnet_ids`:

```hcl
# Before:
subnet_ids = try(each.value.service_type, "Interface") == "Interface" ? distinct(concat(var.subnet_ids, lookup(each.value, "subnet_ids", []))) : null

# After:
subnet_ids = contains(["Interface", "GatewayLoadBalancer"], try(each.value.service_type, "Interface")) ? distinct(concat(try(each.value.service_type, "Interface") == "Interface" ? var.subnet_ids : [], lookup(each.value, "subnet_ids", []))) : null
```

Key behavior:
- `Interface` endpoints: unchanged — `var.subnet_ids` (module-level default) is concatenated with per-endpoint `subnet_ids`
- `GatewayLoadBalancer` endpoints: now receive per-endpoint `subnet_ids` (module-level `var.subnet_ids` is excluded since GWLB endpoints are AZ-specific and should not inherit the shared default)
- `Gateway` endpoints: unchanged — `subnet_ids` remains `null`

## Motivation and Context
`GatewayLoadBalancer` VPC endpoints must be associated with a subnet in order to function. The AWS provider's `aws_vpc_endpoint` resource accepts `subnet_ids` for this type, but the module's conditional logic is silently setting `subnet_ids = null`. This caused GWLB endpoints to be created with no subnet attachment.

This is problematic for centralized inspection architectures where GWLB endpoints are created per-AZ in spoke VPCs and each endpoint must be placed in a specific subnet.

## Breaking Changes
No breaking changes. The only behavioral difference is that `GatewayLoadBalancer` endpoints now correctly receive their per-endpoint `subnet_ids`. Existing configurations that do not use `GatewayLoadBalancer` endpoints are completely unaffected. Existing `GatewayLoadBalancer` endpoints that did not specify `subnet_ids` will continue to receive `null` (via `lookup(each.value, "subnet_ids", [])`).

## How Has This Been Tested?

Tested by deploying GWLB endpoints in a live AWS environment using this module with the following endpoint configuration:

```hcl
endpoints = {
  "gwlb-cac1-1a" = {
    service_name = "com.amazonaws.vpce.ca-central-1.vpce-svc-..."
    service_type = "GatewayLoadBalancer"
    subnet_ids   = ["subnet-abc123"]
  }
  "gwlb-cac1-1b" = {
    service_name = "com.amazonaws.vpce.ca-central-1.vpce-svc-..."
    service_type = "GatewayLoadBalancer"
    subnet_ids   = ["subnet-def456"]
  }
}
```
Before the fix, the endpoints were created but with no subnet association. After the fix, each endpoint is correctly associated with its specified subnet.


modules/vpc-endpoints/main.tf
```diff
@@ -36,7 +36,7 @@ resource "aws_vpc_endpoint" "this" {
   auto_accept       = try(each.value.auto_accept, null)
 
   security_group_ids  = try(each.value.service_type, "Interface") == "Interface" ? length(distinct(concat(local.security_group_ids, lookup(each.value, "security_group_ids", [])))) > 0 ? distinct(concat(local.security_group_ids, lookup(each.value, "security_group_ids", []))) : null : null
-  subnet_ids          = try(each.value.service_type, "Interface") == "Interface" ? distinct(concat(var.subnet_ids, lookup(each.value, "subnet_ids", []))) : null
```
bryantbiggs, line 39: this can simply be:

```suggestion
  subnet_ids          = try(each.value.service_type, "Interface") != "Gateway" ? distinct(concat(var.subnet_ids, lookup(each.value, "subnet_ids", []))) : null
```

codyrat: I added this commit to the PR.

PR #1294, opened 2026-04-05
fix: Prevent empty resources in flow log IAM policy when using external log group

ryotana:
## Description

When `create_flow_log_cloudwatch_log_group = false` and `create_flow_log_cloudwatch_iam_role = true`, the IAM policy's `resources` field becomes an empty list because `local.flow_log_group_arns` iterates over `aws_cloudwatch_log_group.flow_log`.

This fix falls back to `var.flow_log_destination_arn` (with `:*` suffix) when the module does not create the CloudWatch log group, ensuring the IAM policy contains a valid resource ARN.

## Motivation and Context

When users manage their own CloudWatch log group externally and pass it via `flow_log_destination_arn`, the IAM policy for the flow log role ends up with an empty `resources` list, causing `terraform apply` to fail with an AWS API error.

## Breaking Changes

No breaking changes. Existing configurations are unaffected:

- `create_flow_log_cloudwatch_log_group = true`: Uses the existing `for` expression (unchanged path)
- `create_flow_log_cloudwatch_iam_role = false`: IAM policy is not created, `flow_log_group_arns` is never evaluated
- `flow_log_destination_type = "s3"`: IAM policy is not created

## How Has This Been Tested?
I have tested the same in one of my test AWS account and have verified that the changes does change the IAM Policy and VPC flow log's IAM role.

- [x] I have executed `pre-commit run -a` on my pull request (in docker)


bryantbiggs:
image

PR #1279, opened 2026-02-12
fix: Updating AZ regex condition to support EU Sovereign Cloud AZ IDs.

kgoutham93:
## Description

AWS EU Sovereign cloud has different conventions for zone name and zone ids. Instead of asserting the regex on entire AZ id or AZ name, we resort to a simple logic -- split the az by hypen (`-`), if `length(splits) == 2` we assume its AZ id.  

```
{
    "AvailabilityZones": [
        {
            "OptInStatus": "opt-in-not-required",
            "Messages": [],
            "RegionName": "eusc-de-east-1",
            "ZoneName": "eusc-de-east-1a",
            "ZoneId": "euscdee1-az1",
            "GroupName": "eusc-de-east-1-zg-1",
            "NetworkBorderGroup": "eusc-de-east-1",
            "ZoneType": "availability-zone",
            "GroupLongName": "EU (Germany) 1",
            "State": "available"
        }
 ...
}
``` 

## Motivation and Context

AWS EUSC has different naming conventions for AZ name and id.

## Breaking Changes

None, I can think off.

## How Has This Been Tested?
- [x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [ ] I have tested and validated these changes using one or more of the provided `examples/*` projects

I executed examples/simple for `us-east-1` and `eusc-de-east-1` by passing a mix of AZ names and AZ ids to `azs` list. 

- [x] I have executed `pre-commit run -a` on my pull request

PR #1301, opened 2026-07-20
fix: Detect AZ IDs correctly for regions with multi-character prefixes (e.g. eusc-)

mlopezr:
## Description

The module distinguishes an availability-zone name (`eu-central-1a`) from an availability-zone ID (`euc1-az1`) so it can populate either `availability_zone` or `availability_zone_id` on each `aws_subnet`. That detection uses the regex `^[a-z]{2}-`:

```hcl
availability_zone    = length(regexall("^[a-z]{2}-", element(var.azs, count.index))) > 0  ? element(var.azs, count.index) : null
availability_zone_id = length(regexall("^[a-z]{2}-", element(var.azs, count.index))) == 0 ? element(var.azs, count.index) : null
```

The `{2}` hardcodes an assumption that every region prefix is exactly two letters (`us-`, `eu-`, `ap-`, `ca-`, `sa-`, and so on). That assumption is now wrong.

The AWS European Sovereign Cloud region `eusc-de-east-1` has a four-letter prefix. Its zone names look like `eusc-de-east-1a` / `eusc-de-east-1b`, which do not match `^[a-z]{2}-`. The module therefore concludes these are zone IDs and passes them to the EC2 API as `availability_zone_id`, which fails:

```
Error: creating EC2 Subnet: operation error EC2: CreateSubnet, https response error StatusCode: 400, InvalidParameterValue: Value (eusc-de-east-1a) for parameter availabilityZoneId is invalid. Subnets can currently only be created in the following availability zones: euscdee1-az1, euscdee1-az2.
```

This makes the module unusable in `eusc-de-east-1` when passing zone names (the common case, e.g. `azs = slice(data.aws_availability_zones.available.names, 0, 2)`).

## Fix

Widen the region-prefix match from exactly two letters to one-or-more letters:

```diff
-availability_zone    = length(regexall("^[a-z]{2}-", element(var.azs, count.index))) > 0  ? ...
-availability_zone_id = length(regexall("^[a-z]{2}-", element(var.azs, count.index))) == 0 ? ...
+availability_zone    = length(regexall("^[a-z]+-", element(var.azs, count.index))) > 0  ? ...
+availability_zone_id = length(regexall("^[a-z]+-", element(var.azs, count.index))) == 0 ? ...
```

Applied to all 6 subnet resources (`public`, `private`, `database`, `redshift`, `elasticache`, `intra`).

`^[a-z]+-` is correct for both names and IDs. A zone ID always contains a digit before its first hyphen (the region short-code: `use1-`, `usw2-`, `euscdee1-`, `apne1-`, `usw2-lax1-`, `usw2-wl1-...`), so it can never match `^[a-z]+-` (letters-then-hyphen). A zone name always starts with the region name, which is `<letters>-<letters/number>...`, so it always matches. Verified against every zone shape:

| Input                                  | Type | `^[a-z]{2}-` (old)  | `^[a-z]+-` (new) |
| -------------------------------------- | ---- | ------------------- | ---------------- |
| `us-east-1a`                           | name | match → name        | match → name     |
| `eu-central-1a`                        | name | match → name        | match → name     |
| `eusc-de-east-1a`                      | name | no match → ID (BUG) | match → name     |
| `us-west-2-lax-1a` (Local Zone)        | name | match → name        | match → name     |
| `us-east-1-wl1-atl-wlz-1` (Wavelength) | name | match → name        | match → name     |
| `use1-az1`                             | id   | no match → id       | no match → id    |
| `euscdee1-az1`                         | id   | no match → id       | no match → id    |
| `usw2-lax1-az1` (Local Zone id)        | id   | no match → id       | no match → id    |
| `use1-wl1-atl1-wlz1` (Wavelength id)   | id   | no match → id       | no match → id    |

## Testing

- Reproduced the original failure in `eusc-de-east-1` (subnet creation rejected with `InvalidParameterValue` on `availabilityZoneId`).
- With the fix, passing zone names (`eusc-de-east-1a/b`) correctly sets `availability_zone` and subnets create successfully; passing zone IDs (`euscdee1-az1/2`) still correctly sets `availability_zone_id`.
- `terraform fmt` clean.

## Reproduction

```hcl
# In eusc-de-east-1 (AWS European Sovereign Cloud), provider aws ~> 6.0
module "vpc" {
  version = "5.21.0"   # and current master

  name            = "repro"
  cidr            = "10.0.0.0/16"
  azs             = ["eusc-de-east-1a", "eusc-de-east-1b"]  # zone NAMES
  private_subnets = ["10.0.0.0/19", "10.0.32.0/19"]
}
# => Error: InvalidParameterValue: Value (eusc-de-east-1a) for parameter
#    availabilityZoneId is invalid.
```

Closes #1274
Closes #1279


bryantbiggs:
xref #1274 / #1279

mlopezr:
Thanks @bryantbiggs for the referencing those PRs. From what I can see, both were auto-closed by the stale robot.

Also both had issues:
#1274: `-az[0-9]+$` mis-detects Wavelength zone IDs (use1-wl1-atl1-wlz1)
#1279: `split=="2"` mis-detects both Local Zone IDs (usw2-lax1-az1, 3 parts) and Wavelength IDs

Consider this a fresh PR with a smaller, more correct diff so the fix doesn't get lost again.

cc @bpauwels / @kgoutham93 since you both hit this.

mlopezr:
Can any reviewer help please?
Thanks!
cc @bryantbiggs @antonbabenko

bryantbiggs (approved):

antonbabenko:
This PR is included in version 6.7.1 :tada:

PR #1304, opened 2026-08-13
feat: Add per-index Name overrides for NAT gateways and their EIPs

3rwww1:
## Description

Adds `nat_gateway_names` and `nat_eip_names`, mirroring the existing `public_subnet_names` / `private_subnet_names` inputs: explicit per-index `Name` tag values, with the generated name as fallback.

`nat_gateway_tags` and `nat_eip_tags` are single maps merged into every instance, so setting `Name` through them names *all* NAT gateways identically. There is currently no way to give each NAT gateway or EIP its own name.

Our case: we name resources with compressed region codes (`vpc-euw3a` rather than `vpc-eu-west-3a`). Subnets can already do this via `public_subnet_names`; NAT gateways and EIPs cannot.

## Motivation and Context

Implementation matches `public_subnet_names` exactly, including index semantics under `single_nat_gateway`:

```hcl
"Name" = try(
  var.nat_gateway_names[var.single_nat_gateway ? 0 : count.index],
  format("${var.name}-%s", element(var.azs, var.single_nat_gateway ? 0 : count.index)),
)
```

## Breaking Changes

None. Both inputs default to `[]`, and `try()` falls back to the generated name, so existing configurations plan identically.

## How Has This Been Tested?

`examples/complete` now sets `nat_gateway_names` and deliberately omits `nat_eip_names`, so it exercises both the override and the fallback in one place. It plans as `aws_nat_gateway.this[0]` = `NAT Gateway One`, `aws_eip.nat[0]` = `<name>-eu-west-3a`.

`pre-commit run -a`: 8/8 hooks pass (`wrappers/main.tf` is regenerated output from `terraform_wrapper_module_for_each`).

Verified separately with a two-instance plan (both NAT gateways created, `single_nat_gateway = false`):

| Instance | `Name` |
|---|---|
| `with_names` `aws_nat_gateway.this[0]` / `[1]` | `incubator-euw3a` / `incubator-euw3b` |
| `with_names` `aws_eip.nat[0]` / `[1]` | `incubator-eip-euw3a` / `incubator-eip-euw3b` |
| `without_names` `aws_nat_gateway.this[0]` / `[1]` | `legacy-eu-west-3a` / `legacy-eu-west-3b` |
| `without_names` `aws_eip.nat[0]` / `[1]` | `legacy-eu-west-3a` / `legacy-eu-west-3b` |

The `without_names` rows are the unchanged upstream behavior. A partially-filled list also degrades per index rather than erroring: `nat_gateway_names = ["incubator-euw3a"]` yields `incubator-euw3a` and `incubator-eu-west-3b`.


3rwww1:
Can anyone please review this? Thanks!
cc @bryantbiggs @antonbabenko

PR #1305, opened 2026-08-21
feat: Support skip_destroy and deletion protection for flow log log groups

sarahhodne:
## Description

Adds these two variables to the `flow-log` module:

- `cloudwatch_log_group_skip_destroy`: Maps to `aws_cloudwatch_log_group.skip_destroy`
- `cloudwatch_log_group_deletion_protection_enabled`: Maps to `aws_cloudwatch_log_group.deletion_protection_enabled`

I put them both in the same pull request because they feel closely related to me, but I am happy to split it into two pull requests if you'd prefer separate ones.

## Motivation and Context

We are migrating from the VPC flow logs in the base module to the `flow-log` module, and I noticed that `skip_destroy` is not supported there currently. This means that if in the future we want to change an attribute that forces log group recreation (such as storage class or aggregation time), but we want to keep the existing data, we'd need to manually run `terraform state rm` for the existing log group, which we try to avoid with our CI-based apply workflow.

Using a `removed` block doesn't work, because the resource still exists in the configuration, you get an error like this:

```
╷
│ Error: Removed resource still exists
│
│   on ../../modules/flow-log/main.tf line 83:
│   83: resource "aws_cloudwatch_log_group" "this" {
│
│ This statement declares that module.flow_log.aws_cloudwatch_log_group.this was removed, but it is still declared in configuration.
╵
```

`skip_destroy` resolves this, as it means the old group will be left behind.

Since we do not want the log groups to be destroyed, enabling deletion protection also adds a second layer of protection, and also protects against accidental deletion in the web console.

## Breaking Changes

No breaking changes.

## How Has This Been Tested?

  I did not commit an example using `skip_destroy`, since this would leak resources when people use the examples to test.

  How I tested:

  1. Run `terraform apply` in `examples/flow-log` without my changes (i.e. with `master` checked out).
     - The only change I had was changing the region to `eu-west-2`, because the test AWS account I have access to had hit the limit of VPCs in `eu-west-1`.
  3. Apply my code changes, as well as a new example looking like this (which I did not commit to avoid having an example that leaks resources):
     ```hcl
     module "flow_log_deletion_protection" {
       source = "../../modules/flow-log"
    
       name   = local.name
       vpc_id = module.vpc.vpc_id
    
       cloudwatch_log_group_skip_destroy                = true
       cloudwatch_log_group_deletion_protection_enabled = true
    
       tags = local.tags
     }
     ```
  4. Run `terraform apply`
     - The only change is the addition of the new `module.flow_log_deletion_protection` resources.
  6. Add `cloudwatch_log_group_class = "INFREQUENT_ACCESS"` to `module.flow_log_deletion_protection` to force recreation of the CloudWatch log group.
  7. Run `terraform apply`
     - The apply did not error
     - The new CloudWatch log group was created
     - The old one remains in the AWS account (checked in the AWS CloudWatch console)
  9. Set `cloudwatch_log_group_deletion_protection_enabled = true` on `module.flow_log`.
  10. Run `terraform apply`
      - The apply did not error
      - Deletion protection is now enabled on the CloudWatch log group
  12. Run `terraform destroy`
      - Deleting the CloudWatch log group associated with `module.flow_log` failed, due to deletion protection being enabled
  14. Disable deletion protection in the AWS console (to avoid having to re-run `terraform apply` and re-creating all the resources that were just deleted).
  15. Run `terraform destroy` again
      - No errors were reported
      - Checking the CloudWatch web console, I see two log groups remain, both are the ones that were associated with `module.flow_log_deletion_protection`.
  ```
  ❯ pre-commit run --from-ref origin/master --to-ref sarah/flow-logs-skip-destroy-deletion-protection
  Terraform fmt............................................................Passed
  Terraform wrapper with for_each in module................................Passed
  Terraform docs...........................................................Passed
  Terraform validate with tflint...........................................Passed
  Terraform validate.......................................................Passed
  check for merge conflicts................................................Passed
  fix end of files.........................................................Passed
  trim trailing whitespace.................................................Passed
  ```


bryantbiggs (approved):

antonbabenko:
This PR is included in version 6.7.0 :tada:

PR #1306, opened 2026-08-27
fix: Stop apply failing on DNS64 without IPv6 and on empty route table lists

### LINKED ISSUE #944: "Error in function call" when using neither private subnets nor NAT
## Description

TF errors out when trying to create a VPC with database subnets but no private subnets and no NAT gateways:
`Call to function "coalescelist" failed: no non-null arguments`.

It seems that the `aws_route_table.private` isn't created in this case (per `count` condition), but it is apparently meant to be referenced in `aws_route_table_association.database`.

## Versions

- Module version [Required]: 5.0.0

- Terraform version: 1.4.6
- Provider version(s): 5.0.1

```
module "vpc" {
  source = "github.com/terraform-aws-modules/terraform-aws-vpc?ref=v5.0.0"

  name = "foo"
  cidr = "10.0.0.0/16"
  azs  = ["eu-central-1a", "eu-central-1b"]

  enable_dns_hostnames = true
  enable_dns_support   = true
  enable_nat_gateway   = false
  enable_ipv6          = false

  public_subnets = [ "10.0.0.0/24", "10.0.1.0/24" ]
  database_subnets = [ "10.0.10.0/24", "10.0.11.0/24" ]
  #private_subnets = [ "10.0.20.0/24", "10.0.21.0/24" ]
}
```

## Expected behavior

No error.

## Actual behavior

Error in function call (see below).

### Terminal Output Screenshot(s)

```
│ Error: Error in function call
│ 
│   on .terraform/modules/vpc/main.tf line 410, in resource "aws_route_table_association" "database":
│  410:     coalescelist(aws_route_table.database[*].id, aws_route_table.private[*].id),
│     ├────────────────
│     │ while calling coalescelist(vals...)
│     │ aws_route_table.database is empty tuple
│     │ aws_route_table.private is empty tuple
│ 
│ Call to function "coalescelist" failed: no non-null arguments.
```

## Additional context

This configuration worked fine with old version v3.2.0.

When patching the module's `main.tf` like this ...
```
 # There are as many routing tables as the number of NAT gateways
 resource "aws_route_table" "private" {
-  count = local.create_private_subnets && local.max_subnet_length > 0 ? local.nat_gateway_count : 0
+  count = local.create_private_subnets || local.max_subnet_length > 0 ? local.nat_gateway_count : 0
```
... it works correctly. I would submit a PR, but I don't understand the logic / intention well enough to be certain that this is a proper fix for all valid configuration options.

bryantbiggs:
## Description

Three bugs that reproduce on `master`, found by applying every example:

- Subnets set `enable_dns64` and the AAAA record option from `enable_ipv6` alone, but a
  subnet only has an IPv6 CIDR when prefixes are supplied. AWS rejects the modification, so
  `enable_ipv6 = true` with no prefixes fails at apply.
- The database, redshift and elasticache route table associations call `coalescelist` with a
  fallback that can be empty, which raises. Resolves #944
- The IPAM example could never be planned in one pass: it derived subnet CIDRs from
  `aws_vpc_ipam_preview_next_cidr`, a computed attribute, so every subnet `count` failed.

The flow log example also had a permanent diff, because AWS adds its own delivery statement
to the destination bucket policy and the s3-bucket module owns that policy.

Example coverage goes from 67 of 236 root variables to 236 of 236, and from 20 of 51
sub-module variables to 51 of 51. Nine of the fourteen network ACL rule variables had nothing
exercising them at all.

## Motivation and Context

`terraform validate` cannot see any of these; only applying the examples does.

## Breaking Changes

None. The Terraform floor stays at `>= 1.0`.

## How Has This Been Tested?

Every example applied, re-planned clean and destroyed.


antonbabenko (approved):
LGTM! Thanks for finding time to look and this module in particular. 👍

bryantbiggs:
trying to find a bit of time to throw some claude at some of these repos that need a bit more attention - working on it!

antonbabenko:
This PR is included in version 6.7.2 :tada:

PR #1308, opened 2026-08-29
fix: Add the NAT64 route public subnets need for DNS64 to work

### LINKED ISSUE #972: Confusing DNS64 behaviour with public subnets
## Description

`public_subnet_enable_dns64` is `true` by default and with `enable_ipv6 = true;` this enables DNS64 for created public subnets.

However, if a given domain doesn't have an IPv6 record, it's resolved to `64:ff9b::/96` which in turn needs an additional route setup 
to work correctly.

Somewhat similar to #923, but for public subnets.

## Versions

- Module version [Required]: `5.1.1`

- Terraform version: `1.5.4`

- Provider version(s): AWS - `5.10.0`

```
module "aws_vpc" {
  version = "5.1.1"

  name = local.vpc_name
  cidr = var.vpc_cidr
  azs  = local.azs

  public_subnets  = ["10.0.11.0/24", "10.0.12.0/24", "10.0.13.0/24"]

  enable_ipv6 = true
  assign_ipv6_address_on_creation = true
  enable_dns_hostnames = true
  enable_dns_support = true
  map_public_ip_on_launch = true
  public_subnet_assign_ipv6_address_on_creation = true

  enable_dhcp_options = true
  dhcp_options_domain_name = var.domain_name
  dhcp_options_domain_name_servers = ["AmazonProvidedDNS"]

  public_subnet_ipv6_prefixes  = [0, 1, 2]
}
```

`curl -6 api.github.com` from the EC2 instance within one of the public subnets above hangs.

## Expected behavior

DNS64 enabled along with the corresponding route for resolving `64:ff9b::/96`

## Actual behavior

An IP from `64:ff9b::/96` is not routed correctly:(

bryantbiggs:
## Description

DNS64 makes the Amazon resolver synthesise IPv6 addresses inside `64:ff9b::/96` for IPv4
only destinations. Those addresses are unreachable without a route sending that prefix to a
NAT gateway, which performs the translation.

`public_subnet_enable_dns64` defaults to `true`, but no NAT64 route was ever created for
public subnets. The private and database tiers already have one. So on an IPv6 only public
subnet, names resolve and connections then go nowhere, and nothing about the plan or the
apply looks wrong.

This adds the route, on the same conditions the private tier already uses: IPv6 enabled,
the tier has IPv6 prefixes or is IPv6 native, DNS64 enabled for the tier, and a NAT gateway
exists to do the translating.

## Motivation and Context

Fixes #972.

## Breaking Changes

**None.** `variables.tf` is untouched, so no default moves.

A caller without IPv6 sees no change at all. A caller who already has IPv6 public subnets
with DNS64 gains the route that was missing, rather than losing a setting they asked for.

Verified by plan: a VPC with IPv6 public and private subnets and a NAT gateway gains the
new public NAT64 route, and a VPC with no IPv6 gains nothing.

## How Has This Been Tested?

  Run as `LC_ALL=C pre-commit run -a`, fully passing.

  Not applied against live credentials.

## Two related findings, deliberately not fixed here

Both would change behaviour, so they do not belong in a non breaking change.

**Five tiers enable DNS64 by default with no NAT64 route.** `public` is fixed here because
it has a NAT gateway available. `redshift`, `elasticache`, `intra` and `outpost` also
default `*_subnet_enable_dns64` to `true` and have no NAT64 route, and `intra` has no
egress path at all by design, so there is nothing to route to. Making those correct means
either changing defaults or adding routes to tiers that have none today.

**The database NAT64 route is gated on the private tier's settings.** Its `count` reads
`private_subnet_ipv6_prefixes`, `private_subnet_ipv6_native` and
`private_subnet_enable_dns64`, while the database subnet's own `enable_dns64` reads the
database equivalents. Enabling DNS64 on database subnets alone therefore gives them no
route. Correcting it would remove routes for anyone relying on the current behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,7 @@ No modules.
 | [aws_route.private_dns64_nat_gateway](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route) | resource |
 | [aws_route.private_ipv6_egress](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route) | resource |
 | [aws_route.private_nat_gateway](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route) | resource |
+| [aws_route.public_dns64_nat_gateway](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route) | resource |
 | [aws_route.public_internet_gateway](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route) | resource |
 | [aws_route.public_internet_gateway_ipv6](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route) | resource |
 | [aws_route_table.database](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route_table) | resource |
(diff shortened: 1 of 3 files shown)


bryantbiggs:
Closing. The AWS DNS64 and NAT64 documentation places the `64:ff9b::/96` route on the route table of the subnet containing the IPv6-only workloads, and states that subnet should be separate from the one holding the NAT gateway:

> You should first configure a NAT gateway in a public subnet (separate from the subnet containing the IPv6-only workloads).
> Add the following three routes to the route table of the subnet containing the IPv6-only workloads [...] `64:ff9b::/96` route pointing to the NAT gateway.

This change added that route to the public route table, pointing at a NAT gateway in that same public subnet, which the documentation does not describe. The module already does the documented thing for the private and database tiers.

What #972 exposes is that `public_subnet_enable_dns64` defaults to `true` on the one tier that is not intended to hold IPv6-only workloads. Correcting that is a breaking change, so it belongs in the next major version rather than here. #972 stays open.

Reference: https://docs.aws.amazon.com/vpc/latest/userguide/nat-gateway-nat64-dns64.html

